PR #9, opened 2019-11-04
Boilerplate install script

IvanGrginovInf:


package.json
```diff
@@ -77,6 +74,12 @@
     "webpack-manifest-plugin": "^2.0.4",
     "webpack-merge": "^4.2.2"
   },
+  "devDependencies": {
```
iruzevic: please remove this part and aeverything that you added in root package.json. This file is used to fetch all packages for project. I will fix this part

IvanGrginovInf: @iruzevic yup, forgot to remove that, all gone now. 👍

PR #39, opened 2019-12-14
Docpress

dingo-d:
I've added the docpress and started moving the documentation from wiki to there. The build is simple, just `docpress s` and you can see how it looks by going to `http://localhost:3000`.

The documentation needs heavy work because a lot of it is a bit outdated.


iruzevic:
Awesome tnx.

I will merge this to my storybook branch and make all the documentation pages next week. Tnx denis

dingo-d:
@infinum/eightshift-team Feel free to look at this PR 🙂 

I've done porting all the docs, and expanding it a bit. We need to see how we can incorporate sass docs with this (docpress [can be styled](http://docpress.github.io/customization/css.html)).

Not sure if I missed anything.

docs/installation/instalation-and-requirements.md
```diff
@@ -0,0 +1,233 @@
...
+    composer require infinum/eightshift-libs
+    ```
+
+2. If you have an existing Webpack config, look at how the config is set up in [Eightshift Boilerplate repo](https://github.com/infinum/eightshift-boilerplate/blob/develop/webpack.config.js) and merge any changes you have.
+
+The `eightshift-libs` composer package contains a lot of useful abstract classes, interfaces and dependency injection container implementation that you can use in your project. 
+
+### Examples
+
+Inside your theme's `functions.php` you'd register a main class like this
```
IvanGrginovInf: This whole section below doesn't really make sense for `frontend-libs`. You're talking about registering a main class and adding new classes (`php` stuff) which is https://github.com/infinum/eightshift-libs territory. If anything, this should / could be added there.

docs/installation/register-webpack.md
```diff
@@ -0,0 +1,42 @@
+# Register Webpack
+
+In order to use libraries webpack you'll need to have a `webpack.config.js` and `webpack-project.config.js` files in the root of your theme.
```
IvanGrginovInf:
```suggestion
In order to use this library, you'll need to have a `webpack.config.js` and `webpack-project.config.js` files in the root of your theme.
```

docs/installation/instalation-and-requirements.md
dingo-d: Yeah, the issue is that blocks just won't work the way it's described here unless you install `eightshift-libs`. It's kinda a hard dependency. 

You can use it for the mixins and stuff I guess, we could maybe add some notification above about this :+1:

IvanGrginovInf: > Yeah, the issue is that blocks just won't work the way it's described here unless you install eightshift-libs. It's kinda a hard dependency.
> 
> You can use it for the mixins and stuff I guess, we could maybe add some notification above about this 👍

Yeah I'm quite aware, tho that should probably be at the top of `README.md`.

I'm realising the issue that we have:

This whole part should go to https://github.com/infinum/eightshift-libs (for which we also need docs). But it's also important for this lib. And ideally we need all of it in one place.

So we probably need to think about having a separate DocPress repo which will hold the stuff for both `eightshift-libs` and `eightshift-frontend-libs` (rather than having each repository have it's own DocPress OR even worse, this repository holding docs for both).

For now I guess it's fine but I'd rather if we solved this sooner rather than later.

iruzevic:
I have merger your branch to my and added changes so I will close this PR

PR #51, opened 2020-01-03
Hotfix/brosersync override

tihomirselak:
If value of key was false, browserSync would not be overridden, if we would like add possibility for value to be false, we should only check if key is present

### DIFF
--- a/webpack/helpers.js
+++ b/webpack/helpers.js
@@ -70,7 +70,7 @@ function getConfig(projectDir, proxyUrl, projectPathConfig, assetsPathConfig, ou
 }
 
 /**
- * Check if user config is added and it is used.s
+ * Check if user config is added and it is used
  *
  * @param {object} config Config object to check.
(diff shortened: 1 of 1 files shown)


webpack/helpers.js
```diff
@@ -70,15 +70,15 @@ function getConfig(projectDir, proxyUrl, projectPathConfig, assetsPathConfig, ou
...
+ * Check if user config is added and it is used
  *
  * @param {object} config Config object to check.
  * @param {string} key Config object key to test if is false.
  *
  * @since 2.0.0
  */
 function isUsed(config, key) {
-  if (config.hasOwnProperty(key) && config[key]) {
+  if (config.hasOwnProperty(key)) {
```
dingo-d: Hmmm but we do need to check if the property is `true/false`, no?

tihomirselak: closed

PR #59, opened 2020-01-09
Feature/layout components

iruzevic:
Moved all src/blocks/layout/... to src/blocks/components for consistency and to be more semantically correct. Later one everything in WP will be block so we are prepared.


iruzevic:
> Will this have any impact on how boilerplate handles these components/layouts?
> 
> We should watch the core blog about GB development to see how this will pan out in the future when full site editing will be available.

@IvanGrginovInf and I agreed that everything that is not block will be a component an @IvanGrginovInf will make a header fully modular with options for multiple mobile menues so we can speed up development

PR #60, opened 2020-01-14
added new divider block

iruzevic:
## Added
- New divider block

### DIFF
--- a/blocks/init/custom/divider/divider-style.scss
+++ b/blocks/init/custom/divider/divider-style.scss
@@ -1,0 +1,15 @@
+$block-divider: (
+  colors: (
+    primary: $base-primary-color,
+    black: $base-black-color,
+    white: $base-white-color,
+  ),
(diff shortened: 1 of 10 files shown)


blocks/init/custom/divider/divider-style.scss
```diff
@@ -0,0 +1,15 @@
+$block-divider: (
+  colors: (
+    primary: $base-primary-color,
+    black: $base-black-color,
+    white: $base-white-color,
+  ),
+);
+
+.block-divider {
+  height: 1px;
```
IvanGrginovInf: Is this what determines the divider's height? I.e. it's always `1px` high? If so it should probably be defined in block's options.

iruzevic: Yes, height depends on the project but in my experience it is in 99% cases 1px so lets not complicate this component

PR #62, opened 2020-01-15
Underline and underline text mixins

volfkarlo:
Mixins that I've announced on WP standup are done.

Please pay attention to explanation and way it's written. I assume I will have a lot of mistakes. To me it makes sense but it's quite hard for me to put myself in someone else's shoes.

### Changelog
- Added underline-text mixin
- Added underline mixin
- Added for-each-attribute mixin


styles/scss/mixins/_underline-text.scss
```diff
@@ -0,0 +1,109 @@
...
+/// @access public
+/// @author Karlo Volf
+/// @param {percentage} $thickness [10%] - Thickness of the line in percentage.
+/// @param {seconds} $duration [.5s] - How long animation lasts.
+/// @param {timing-function} $timing-function [cubic-bezier(.79, .01, .22, .99)] - Execution speed function.
+/// @param {hex} $color [currentColor] - Color of the line.
+/// @param {boolean} $wrapper [false] - If wrapper is used as hover reference.
+/// @param {string} $element [.underline-text] - Element on which you want underline.
+/// @param {string} $state [hover] - The state in which it becomes underline element.
+/// @example
```
iruzevic: add spacing here

styles/scss/mixins/_underline-text.scss
```diff
@@ -0,0 +1,109 @@
...
+/// .test {
+///   @include underline-text();
+/// }
+/// .test2 {
+///   @include underline-text(10, .7s, ease-in, #132031, true, a.link, hover);
+/// }
+/// @output
```
iruzevic: add spacing here

styles/scss/mixins/_underline-text.scss
```diff
@@ -0,0 +1,109 @@
...
+///   background-repeat: no-repeat;
+///   background-position-y: bottom;
+///   background-size: 0% 100%;
+///   &:hover {
+///     background-position-x: right;
+///     background-position-y: bottom;
+///     background-size: 100% 100%;
+///   }
+/// }
+/// .test2 {
```
iruzevic: spacing

styles/scss/mixins/_underline-text.scss
```diff
@@ -0,0 +1,109 @@
...
+/// @param {seconds} $duration [.5s] - How long animation lasts.
+/// @param {timing-function} $timing-function [cubic-bezier(.79, .01, .22, .99)] - Execution speed function.
+/// @param {hex} $color [currentColor] - Color of the line.
+/// @param {boolean} $wrapper [false] - If wrapper is used as hover reference.
+/// @param {string} $element [.underline-text] - Element on which you want underline.
+/// @param {string} $state [hover] - The state in which it becomes underline element.
+/// @example
+/// .test {
+///   @include underline-text();
+/// }
```
iruzevic: spacing

styles/scss/mixins/_underline-text.scss
```diff
@@ -0,0 +1,109 @@
...
+
+  @else {
+    @include for-each-attribute($underline-inactive-state);
+
+    &:#{$state} {
+      @include for-each-attribute($underline-active-state);
+    }
+  }
+}
+@mixin for-each-attribute($map) {
```
iruzevic: Hm I think this can be extracted in a separate standalone mixin @gabriel-glo any thoughts?

styles/scss/mixins/_underline.scss
```diff
@@ -0,0 +1,109 @@
+/// Mixin that will underline blocks elements. Each parameter is optional.
+/// Breakpoint is manually added with last parameter.
+///
+/// @access public
+/// @author Karlo Volf
```
iruzevic: spacing

styles/scss/mixins/_underline-text.scss
```diff
@@ -0,0 +1,109 @@
...
+///  First is to include in the class you want to hover over while text
+///  you want to underline needs to have attached class .underline-text
+///  or you can send your custom class name via parameter $element.
+///  Also you have to set $wrapper parameter to true.
+///  Second is to just include it in whichever element you want. In this
+///  case, parameter $wrapper should stay unchanged.
+///  Other parameters are optional and they allow you customize your transition.
+///
+/// @access public
+/// @author Karlo Volf
```
iruzevic: spacing

styles/scss/mixins/_underline.scss
```diff
@@ -0,0 +1,109 @@
...
+/// @param {pixel} $thickness [2px] - Thickness of the line.
+/// @param {seconds} $duration [.3s] - How long animation lasts.
+/// @param {timing-function} $timing-function [cubic-bezier(.79, .01, .22, .99)] - Execution speed function.
+/// @param {hex} $color [currentColor] - Color of the line.
+/// @param {pixel} $breakpoint [1080px] - Minimum width of the element.
+/// @example
```
iruzevic: spacing

styles/scss/mixins/_underline.scss
```diff
@@ -0,0 +1,109 @@
...
+/// .test {
+///   @include underline();
+/// }
```
iruzevic: spacing

styles/scss/mixins/_underline.scss
```diff
@@ -0,0 +1,109 @@
...
+/// .test2 {
+///   @include underline(4px, 0.5s, ease-in, #132031, 300px);
+/// }
```
iruzevic: spacing

styles/scss/mixins/_underline.scss
```diff
@@ -0,0 +1,109 @@
...
+///     background-color: currentColor;
+///   }
+///   @media (min-width: 1080px) {
+///     .test:hover::after {
+///       content: "";
+///       transform: scaleX(1);
+///       transform-origin: right center;
+///     }
+///   }
+/// }
```
iruzevic: spacing

styles/scss/mixins/_underline.scss
```diff
@@ -0,0 +1,109 @@
...
+
+  @media (min-width: $breakpoint) {
+    &:hover::after {
+      content: '';
+      transform: scaleX(1);
+      transform-origin: right center;
+    }
+  }
+}
+@mixin for-each-attribute($map) {
```
iruzevic: as before

styles/scss/mixins/_underline-text.scss
```diff
@@ -0,0 +1,109 @@
+///  This mixin has 2 possible way of using it.
+///  First is to include in the class you want to hover over while text
+///  you want to underline needs to have attached class .underline-text
+///  or you can send your custom class name via parameter $element.
+///  Also you have to set $wrapper parameter to true.
+///  Second is to just include it in whichever element you want. In this
+///  case, parameter $wrapper should stay unchanged.
+///  Other parameters are optional and they allow you customize your transition.
```
IvanGrginovInf: First of all you can add some empty lines in the description (tho I think it renders the same in the sassdoc itself).

```
///  This mixin has 2 possible way of using it.
///
///  First is to include in the class you want to hover over while text
///  you want to underline needs to have attached class .underline-text
///  or you can send your custom class name via parameter $element.
///
///  Also you have to set $wrapper parameter to true.
///
///  Second is to just include it in whichever element you want. In this
///  case, parameter $wrapper should stay unchanged.
///
///  Other parameters are optional and they allow you customize your transition.
```

Second, the description is markdown so you can highlight things.

```markdown
///  This mixin has 2 possible ways of using it.
///
/// #### Example 1
///
///  Include in the class you want to hover over while text
///  you want to underline needs to have attached class `.underline-text`
///  or you can send your custom class name via parameter `$element`.
///
///  Also you have to set `$wrapper` parameter to true.
///
/// #### Example 2
///  
///  Just include it in whichever element you want. In this
///  case, parameter `$wrapper` should stay unchanged.
///
///  Other parameters are optional and they allow you customize your transition.
```

> Include in the class you want to hover over while text you want to underline needs to have attached class `.underline-text`

The first way's explanation is kinda confusing. I need to include the mixing AND add the class of `underline-text` to the element? 

> or you can send your custom class name via parameter `$element`.

So instead of the both things above, I can just add the class as parameter?

Other than that this is the best thing ever ❤️

styles/scss/mixins/_underline-text.scss
```diff
@@ -0,0 +1,109 @@
...
+///
+/// @access public
+/// @author Karlo Volf
+/// @param {percentage} $thickness [10%] - Thickness of the line in percentage.
```
gabriel-glo: Percentage of what?
Maybe explain how the thickness of the line will be calculated.

styles/scss/mixins/_underline-text.scss
```diff
@@ -0,0 +1,109 @@
...
+/// @param {boolean} $wrapper [false] - If wrapper is used as hover reference.
+/// @param {string} $element [.underline-text] - Element on which you want underline.
+/// @param {string} $state [hover] - The state in which it becomes underline element.
+/// @example
+/// .test {
+///   @include underline-text();
+/// }
+/// .test2 {
+///   @include underline-text(10, .7s, ease-in, #132031, true, a.link, hover);
+/// }
```
gabriel-glo: Maybe add another example of mixing usage. This way some mixin arguments can be used by default and only specific ones will be overridden.
```
@include underline-text(
  $thickness: 10,
  $duration: .7s
  ...
);
```

styles/scss/mixins/_underline-text.scss
gabriel-glo: Definitely a separate mixin. This can be reused on other places.
I'm thinking this could also be turned into a scss function.

styles/scss/mixins/_underline.scss
```diff
@@ -0,0 +1,109 @@
...
+/// @param {hex} $color [currentColor] - Color of the line.
+/// @param {pixel} $breakpoint [1080px] - Minimum width of the element.
+/// @example
+/// .test {
+///   @include underline();
+/// }
+/// .test2 {
+///   @include underline(4px, 0.5s, ease-in, #132031, 300px);
+/// }
+/// @output
```
gabriel-glo: Also add third option too include a mixin, same as my comment above.

styles/scss/mixins/_underline-text.scss
dingo-d: I agree the first example is confusing to me. The text is in a separate class that is wrapped with the class that holds the hover?

Add a comma after `Also`, and `the` before `$wrapper` like

 ```
Also, you have to set the `$wrapper` parameter to true.
```

styles/scss/mixins/_underline-text.scss
volfkarlo: I would say that it can't be a function because of the @return paramater. I didn't find a way to return all of the parameters at once. 
@gabriel-glo do you maybe know how to?

iruzevic:
@volfkarlo da li je ovo gotovo?

volfkarlo:
rekao bih da da.

iruzevic:
@volfkarlo onda još dodaj u change log što si napravio i mergaj to

Issue #42, opened 2019-12-17
Convert all jQuery to Vanilla JS


iruzevic:

kmaric69:
I will close this ticket since we removed jQuery from all blocks.

PR #70, opened 2020-01-22
updating storybook config

iruzevic:
I have changes stuff a little bit. In general there are a lot of changes but only because I have moved blocks from `blocks>init>blocks` to `blocks>init>src>blocks`. This was done so se can have consistant folder structure to a final project. @IvanGrginovInf there will be changes in the rename script just to fix this folder changes

Also there are some changes on storybook config so it can be included in to each project with few easy steps.


.storybook/parts/default-categories.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * Manualy populate categories for blocks. This is generated in the PHP part of the real project.
+ */
+wp.data.dispatch( 'core/blocks').setCategories([
+  {
+    slug: 'eightshift',
+    title: 'Eightshift',
+    'icon': 'admin-settings',
```
dingo-d:
```suggestion
    icon: 'admin-settings',
```

This doesn't have to be quoted?

.storybook/preview-head.html
```diff
@@ -1,3 +1,7 @@
 <script>
   window.wp = {};
 </script>
+
+<!-- REMOVE ONCE jQuery is removed from the libs. -->
```
dingo-d: Do we have an issue for this? Because _TODO_s are often forgotten in the code.

webpack/storybook.js
```diff
@@ -0,0 +1,63 @@
...
+      },
+    ],
+  });
+
+  /**
+   * Extract css from sass.
+   */
+  config.plugins.push(new MiniCssExtractPlugin());
+
+  // REMOVE ONCE jQuery is removed from the libs.
```
dingo-d: Same as with the previous comment

.storybook/preview-head.html
kmaric69: yes, there is an open issue for this:
https://github.com/infinum/eightshift-frontend-libs/issues/42#event-2893662111

webpack/storybook.js
kmaric69: 🆙

.storybook/preview-head.html
dingo-d: Ok, then this comment needs to be removed 🙂

iruzevic:
> Will this enable users to locally run storybooks?
> 
> Also, please don't merge this before create a PR fixing the setup script for the new paths.

Yes it will :D

iruzevic:
Fixed PRs

PR #92, opened 2020-03-16
Scroll to target component

gabriel-glo:
Add scroll-to-target component that works similar to scroll-to-top component

Component has options for defining target of the scroll action and text of the scroll-to link.
It works by getting the `href` attribute and scrolling to element with matching `id` attribute

### DIFF
--- a/blocks/init/src/blocks/components/scroll-to-target/docs/story.js
+++ b/blocks/init/src/blocks/components/scroll-to-target/docs/story.js
@@ -1,0 +1,34 @@
+import React from 'react'; // eslint-disable-line no-unused-vars
+import readme from './readme.md';
+import { ScrollToTargetEditor } from '../components/scroll-to-target-editor';
+import { ScrollToTargetOptions } from '../components/scroll-to-target-options';
+
+export default {
+  title: 'Components|Scroll To Target',
+  parameters: {
+    notes: readme,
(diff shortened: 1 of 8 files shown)


blocks/init/src/blocks/components/scroll-to-target/docs/story.js
```diff
@@ -0,0 +1,34 @@
...
+
+export default {
+  title: 'Components|Scroll To Target',
+  parameters: {
+    notes: readme,
+  },
+};
+
+const editorProps = {
+  scrollText: 'Scroll to target',
```
dingo-d: These are not translatable?

blocks/init/src/blocks/components/scroll-to-target/docs/story.js
```diff
@@ -0,0 +1,34 @@
...
+};
+
+const optionsProps = {
+  scrollText: 'Scroll to target',
```
dingo-d: Same as the above

blocks/init/src/blocks/components/scroll-to-target/docs/story.js
gabriel-glo: I don't think they should be, as these are only example strings for Storybook

PR #97, opened 2020-03-20
Tweak header components

IvanGrginovInf:
Changelog
---
* Added support for `behind` and `top` drawers / mobile menus (slide from top or fade in)
* Better vertical / horizontal centering (more `flex`, less `height`, proper justifying)
* Removed some `height: 100%` which made styling links difficult and coupled to header / footer height
* Fixed a bug with centering where logo wouldn't be quite centered on mobile - seemingly due to having `flex-basis` set to auto. Setting it to any other value (such as `1px`) fixed the issue.
* Fixed a bug where `drawer` wouldn't work if `page-overlay` did not exist

### DIFF
--- a/blocks/init/src/blocks/components/header/header-style.scss
+++ b/blocks/init/src/blocks/components/header/header-style.scss
@@ -33,26 +33,34 @@ $header: (
   }
 
   &__column {
+    display: flex;
     flex: 1 0 auto;
-    height: 100%;
     position: relative;
+
+    &--left {
+      justify-content: left;
+    }
 
     &--center {
       text-align: center;
+      justify-content: center;
     }
 
     &--right {
       text-align: right;
+      justify-content: flex-end;
     }
 
     @include media(desktop up) {
       &--left {
         order: 2;
+        justify-content: center;
       }
 
       &--center {
-        text-align: left;
         order: 1;
+        text-align: left;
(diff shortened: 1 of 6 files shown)


blocks/init/src/blocks/components/header/header-style.scss
```diff
@@ -33,26 +33,34 @@ $header: (
   }
 
   &__column {
-    flex: 1 0 auto;
-    height: 100%;
+    display: flex;
+    flex: 1 0 1px;
```
kmaric69: Why `flex-basis: 1px`?

IvanGrginovInf: I've mentioned in the changelog:

>Fixed a bug with centering where logo wouldn't be quite centered on mobile - seemingly due to having flex-basis set to auto. Setting it to any other value (such as 1px) fixed the issue.

Not sure why this is happening tho.

Most likely specific implementations of these 3 flex columns will have to be tweaked for your project. (for example, as I've noticed while doing our Gutenberg task, it's behaving kinda weird when you have 2 items in a single column - when `flex-basis: 1px` - but it works much better for the default header this way).

IvanGrginovInf:
@iruzevic I've added changelog, is this ok to merge?

PR #98, opened 2020-03-30
Remove some aliases with module imports

IvanGrginovInf:
To do
---
* [x] Test `npm run lintJs`
* [x] Test in Storybook context (working inside `frontend-libs` repository)
* [x] Test in project context (like in `eightshift-boilerplate` when `frontend-libs` in installed inside `node_modules`) - shouldn't be any issues tho (_unless alias clashes with normal module import_)
* [ ] Test in local Storybook (building a Storybook inside your project) - should be the same as the project context one 
* [ ] More research to see if this can be done for other aliases (specifically `.scss` ones and those from `node_modules`) - those are a bit tricky because (especially ones from `node_modules`) because their paths vary depending on if we're in the context of a project (like Eightshift Boilerplate) or in the context of Storybook (working from inside `frontend-libs` repository) - already tried a bunch of things for this but with little luck

Changelog
---
* Replaced some aliases with import from module (for example: `import { ColorPaletteCustom } from '@eightshift/frontend-libs`, instead of `import { ColorPaletteCustom } from 'EightshiftComponentColorPalette`)
  - Had to build an alias for `@eightshift/frontend-libs` in order for this to work while we're working on Storybook from inside the `frontend-libs` repp

### DIFF
--- a/webpack/aliases.js
+++ b/webpack/aliases.js
@@ -5,6 +5,7 @@
  */
 
 const path = require('path');
+const { name } = require('../package.json');
 
 module.exports = (packagesPath) => {
   return {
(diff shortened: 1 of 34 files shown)


webpack/aliases.js
```diff
@@ -11,6 +11,11 @@ module.exports = (packagesPath) => {
     resolve: {
       alias: {
 
+        // Alias this package for use when we're working on Storybook
+        // With this you can always import things from index.jx using:
```
kmaric69, line 15: small typo in index.jx, nothing serious :)

PR #99, opened 2020-04-02
Editor color pallete

IvanGrginovInf:
Changelog
---
* `editor-color-pallete` - Modified the `ColorPaletteCustom` component to get colors by default from WordPress's global store
* `editor-color-pallete` - Modified all uses of `ColorPaletteCustom` component to not override default colors (except for wrapper)
* `editor-color-pallete` - Added a helper (using React hooks) for reading colors from WordPress's global store in `scripts/get-pallete-colors.js`
* Added docs for `editor-color-pallete`

These changes are backwards compatible with existing blocks setting colors "the old way" and nothing should break if you were to update libs on an existing project.

Also instead of using `useSelect` (which isn't supported on WP < 5.3), I've used `withSelect` HOC on `ColorPaletteCustom` which should work everywhere. There's a helper which uses `useSelect` to read colors from WP's global store for when you need to override colors on specific components (but still want to use some of editor-color-pallete colors).

This needs to be merged AFTER PR on `libs` adding editor color pallete - https://github.com/infinum/eightshift-libs/pull/50

_@iruzevic for some reason the `components/color-palette-custom` & `components/heading-level` don't seem to be showing in Storybook, do you happen to know why this is the case?_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ This projects adheres to [Semantic Versioning](https://semver.org/) and [Keep a 
 
 ## [Unreleased]
 * Added `page-overlay` to block copying.
+* `editor-color-pallete` - Modified the `ColorPaletteCustom` component to get colors by default from WordPress's global store
+* `editor-color-pallete` - Modified all uses of `ColorPaletteCustom` component to not override default colors (except for wrapper)
+* `editor-color-pallete` - Added a helper (using React hooks) for reading colors from WordPress's global store.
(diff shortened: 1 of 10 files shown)


IvanGrginovInf:
@iruzevic  This shouldn't have been merged because it's failing the build - meaning the E2E test is failing, i.e. the setup script isn't working now (I need to look into why exactly). 😢

iruzevic:
I havw been fixin it in my branch. It will be fixed today

PR #105, opened 2020-04-22
Feature/loading fix

iruzevic:
### Added
- missing registerBlocks import

### Changed
- js loader to be able to handle components from the lib


components/color-palette-custom/color-palette-custom.js
```diff
@@ -1,5 +1,5 @@
+import React, { useState } from 'react'; // eslint-disable-line no-unused-vars
```
IvanGrginovInf, line 1: So this is because of Storybook? Because you're already importing react in `story.js` here: https://github.com/infinum/eightshift-frontend-libs/blob/1558e1e364f7688a99fb9db90c401fe97cfc2baf/components/color-palette-custom/docs/story.js

That isn't enough?

package.json
```diff
@@ -29,14 +29,14 @@
     "lintStyle": "npm run __stylelintTheme",
     "lintJs": "npm run __eslintTheme",
     "lint": "npm run lintJs && npm run lintStyle",
-    "storybook": "start-storybook",
-    "build-storybook": "build-storybook -c .storybook -o docs/storybook",
```
volfkarlo, line 33: `"storybook": "-storybook -storybook` is ok call?

volfkarlo: Sorry I misunderstood the difference. Ignore this comment.

components/color-palette-custom/color-palette-custom.js
iruzevic: Yes, you have to import it in to every single file that is used inside of storybook

PR #115, opened 2020-05-06
Feature/global variables

iruzevic:
- fixing inline logic from components
- updating better destructuring props
- adding newer version of heading level select component, I just copyed the component from Gutenberg because it is not published yet (https://github.com/WordPress/gutenberg/blob/master/packages/block-library/src/heading/heading-toolbar.js). When it is we will replace it with the official one.

Everything is tested, and this is the last PR for this release


blocks/init/src/blocks/components/button/components/button-options.js
```diff
@@ -30,14 +30,15 @@ export const buttonColors = () => {
...
-      styleSizeWidth,
-      id,
-    },
+    title,
+    url,
+    styleSize,
+    styleColor,
+    styleSizeWidth,
+    id,
+  } = props.button;
```
dingo-d, line 39: This couldn't be written as

```js
export const ButtonOptions = ({button}) => {
  const {
    title,
    url,
    styleSize,
    styleColor,
    styleSizeWidth,
    id,
  } = button;
```

Like in the above example?

blocks/init/src/blocks/components/heading/components/heading-editor.js
```diff
@@ -1,28 +1,30 @@
...
 import { __ } from '@wordpress/i18n';
 import { RichText } from '@wordpress/block-editor';
 
 export const HeadingEditor = (props) => {
+  const {
+    content,
+    styleAlign,
+    styleColor,
+    styleSize,
+  } = props.heading;
```
dingo-d, line 12: Same as the button example, instead of `(props)` -> `({heading})` in the arrow function argument.

blocks/init/src/blocks/components/heading/components/heading-toolbar.js
```diff
@@ -5,10 +5,11 @@ import { HeadingLevel } from '@eightshift/frontend-libs/scripts/components';
 
 export const HeadingToolbar = (props) => {
   const {
-    heading: {
-      level,
-      styleAlign,
-    },
+    level,
+    styleAlign,
+  } = props.heading;
```
dingo-d, line 10: Same

blocks/init/src/blocks/components/heading/components/heading-options.js
```diff
@@ -11,10 +11,11 @@ export const headingSizes = [
 
 export const HeadingOptions = (props) => {
   const {
-    heading: {
-      styleColor,
-      styleSize,
-    },
+    styleColor,
+    styleSize,
+  } = props.heading;
```
dingo-d, line 16: Same

blocks/init/src/blocks/components/image/components/image-editor.js
```diff
@@ -1,50 +1,46 @@
...
 import { Fragment } from '@wordpress/element';
 import { MediaPlaceholder } from '@wordpress/editor';
 import { Button, Dashicon } from '@wordpress/components';
 
 export const ImageEditor = (props) => {
+  const {
+    url,
+    accept = 'image/*',
+    allowedTypes = ['image'],
+  } = props.media;
```
dingo-d, line 12: ⬆️

blocks/init/src/blocks/components/link/components/link-editor.js
```diff
@@ -1,24 +1,26 @@
 import React from 'react'; // eslint-disable-line no-unused-vars
+import classnames from 'classnames';
 import { __ } from '@wordpress/i18n';
 import { RichText } from '@wordpress/block-editor';
 
 export const LinkEditor = (props) => {
+  const {
+    title,
+    styleColor,
+  } = props.link;
```
dingo-d, line 10: ⬆️

blocks/init/src/blocks/components/link/components/link-options.js
```diff
@@ -20,11 +20,12 @@ export const linkColors = () => {
...
   const {
-    link: {
-      url,
-      styleColor,
-      isAnchor,
-    },
+    url,
+    styleColor,
+    isAnchor,
+  } = props.link;
```
dingo-d, line 26: ⬆️

blocks/init/src/blocks/components/lists/components/lists-editor.js
```diff
@@ -1,24 +1,26 @@
 import React from 'react'; // eslint-disable-line no-unused-vars
+import classnames from 'classnames';
 import { __ } from '@wordpress/i18n';
 import { RichText } from '@wordpress/block-editor';
 
 export const ListsEditor = (props) => {
+  const {
+    content,
+    ordered,
+  } = props.lists;
```
dingo-d, line 10: ⬆️

blocks/init/src/blocks/components/paragraph/components/paragraph-editor.js
```diff
@@ -1,33 +1,34 @@
...
 import { __ } from '@wordpress/i18n';
 import { RichText } from '@wordpress/block-editor';
 
 export const ParagraphEditor = (props) => {
+  const {
+    content,
+    styleAlign,
+    styleColor,
+    styleSize,
+  } = props.paragraph;
```
dingo-d, line 12: ⬆️

blocks/init/src/blocks/components/paragraph/components/paragraph-options.js
```diff
@@ -11,45 +11,41 @@ export const paragraphSizes = [
...
 export const ParagraphOptions = (props) => {
   const {
-    paragraph: {
-      styleColor,
-      styleSize,
-      removeStyle,
-    },
+    styleColor,
+    styleSize,
+  } = props.paragraph;
```
dingo-d, line 16: :arrow_up:

blocks/init/src/blocks/components/paragraph/components/paragraph-options.js
```diff
@@ -11,45 +11,41 @@ export const paragraphSizes = [
...
-          }
 
-          {onChangeStyleSize &&
-            <SelectControl
-              label={__('Paragraph Font Size', 'eightshift-boilerplate')}
-              value={styleSize}
-              options={paragraphSizes}
-              onChange={onChangeStyleSize}
-            />
+      {onChangeStyleColor &&
```
dingo-d, line 26: I still didn't get my answer on why do we need `onChangeStyleColor` check here? I'd really love somebody explaining that to me 🙂

blocks/init/src/blocks/components/paragraph/components/paragraph-toolbar.js
```diff
@@ -4,24 +4,20 @@ import { Fragment } from '@wordpress/element';
 
 export const ParagraphToolbar = (props) => {
   const {
-    paragraph: {
-      styleAlign,
-      removeStyle,
-    },
+    styleAlign,
+  } = props.paragraph;
```
dingo-d, line 8: ⬆️

blocks/init/src/blocks/components/paragraph/components/paragraph-toolbar.js
```diff
@@ -4,24 +4,20 @@ import { Fragment } from '@wordpress/element';
...
+
+  const {
     onChangeStyleAlign,
   } = props;
```
dingo-d, line 12: I wonder if this could have all been written like:

```js
export const ParagraphToolbar = ({paragraph, onChangeStyleAlign}) => {
  const { styleAlign } = paragraph;
}
```

?

blocks/init/src/blocks/components/video/components/video-editor.js
```diff
@@ -1,39 +1,28 @@
...
 import { Fragment } from '@wordpress/element';
 import { MediaPlaceholder } from '@wordpress/editor';
 import { Button, Dashicon } from '@wordpress/components';
 
 export const VideoEditor = (props) => {
+  const {
+    url,
+    accept = 'video/*',
+    allowedTypes = ['video'],
+  } = props.media;
```
dingo-d, line 12: Like the above comments

blocks/init/src/blocks/custom/button/button-block.js
```diff
@@ -7,6 +7,7 @@ import { ButtonEditor } from './components/button-editor';
 import { ButtonOptions } from './components/button-options';
 
 export const Button = (props) => {
```
dingo-d, line 9: `{attributes}` instead of props?

blocks/init/src/blocks/custom/card-list/card-list-block.js
```diff
@@ -7,6 +7,7 @@ import { CardListEditor } from './components/card-list-editor';
 import { CardListOptions } from './components/card-list-options';
 
 export const CardList = (props) => {
```
dingo-d, line 9: Like the attributes comment above

blocks/init/src/blocks/custom/cards-grid/cards-grid-block.js
```diff
@@ -4,6 +4,7 @@ import manifest from './manifest.json';
 import { CardsGridEditor } from './components/cards-grid-editor';
 
 export const CardsGrid = (props) => {
```
dingo-d, line 6: attributes comment

blocks/init/src/blocks/custom/cards-list/cards-list-block.js
```diff
@@ -4,6 +4,7 @@ import manifest from './manifest.json';
 import { CardsListEditor } from './components/cards-list-editor';
 
 export const CardsList = (props) => {
```
dingo-d, line 6: `{attributes}`

blocks/init/src/blocks/custom/carousel-image/carousel-image-block.js
```diff
@@ -5,6 +5,7 @@ import { CarouselImageEditor } from './components/carousel-image-editor';
 import './hooks';
 
 export const CarouselImage = (props) => {
```
dingo-d, line 7: ⬆️

blocks/init/src/blocks/custom/carousel/carousel-block.js
```diff
@@ -7,6 +7,7 @@ import { CarouselOptions } from './components/carousel-options';
 import { CarouselEditor } from './components/carousel-editor';
 
 export const Carousel = (props) => {
```
dingo-d, line 9: ⬆️

blocks/init/src/blocks/components/paragraph/paragraph.php
```diff
@@ -14,31 +14,24 @@
...
+$style_color     = isset( $paragraph['styleColor'] ) ? "{$component_class}__color--{$paragraph['styleColor']}" : '';
+$style_size      = isset( $paragraph['styleSize'] ) ? "{$component_class}__size--{$paragraph['styleSize']}" : '';
+
+$paragraph_class = "
+  {$component_class}
+  {$style_color}
+  {$style_align}
+  {$style_size}
+  {$block_class}__paragraph
+";
```
IvanGrginovInf, line 31: We have the equivalent `classNames()` function (from JS) in PHP, you can do it like:

```php
<?php

Components::classnames([
  $component_class,
  $style_color,
  $style_align,
  $style_size,
  "{$block_class}__paragraph"
]);
```

blocks/init/src/blocks/components/video/video.php
```diff
@@ -10,23 +10,17 @@
 
 $media = $attributes['media'] ?? [];
 
-if ( empty( $media['url'] ) ) {
+if ( ! $media['url'] ) {
```
IvanGrginovInf, line 13: won't this throw a warning if `$media['url']` key is undefined?

blocks/init/src/blocks/custom/card-list/card-list.php
```diff
@@ -16,14 +16,9 @@
...
 
-$component_class  = "
-  {$block_class}
-  {$block_class}__media-position--{$media_position}
-";
-
 ?>
 
-<div class="<?php echo esc_attr( $component_class ); ?>">
+<div class="<?php echo esc_attr( "{$block_class} {$block_class}__media-position--{$media_position}" ); ?>">
```
IvanGrginovInf, line 21: If we try to make .php and React block / component templates similar we should follow the same guiding principle / rules.

This should not be done inline but rather a few lines above (similar to how it was before), using `Components::classnames()` preferably (like you'd do in React).

blocks/init/src/blocks/custom/carousel/carousel.php
```diff
@@ -13,16 +13,10 @@
...
-  {$block_class}
-  {$block_js_class}
-  swiper-container
-";
-
 ?>
 
 <div
-  class="<?php echo esc_attr( $component_class ); ?>"
+  class="<?php echo esc_attr( "{$block_class} {$block_js_class} swiper-container" ); ?>"
```
IvanGrginovInf, line 19: ⬆️

blocks/init/src/blocks/components/button/components/button-options.js
iruzevic: No it can'. Because button is a key inside a props object. And we are using props in the line 48. I will ask JS team about this.

iruzevic: I have asked https://infinum.slack.com/archives/C027S03F6/p1588754084207700

blocks/init/src/blocks/components/paragraph/paragraph.php
iruzevic: Tnx

blocks/init/src/blocks/components/video/video.php
iruzevic: tnx

blocks/init/src/blocks/custom/card-list/card-list.php
iruzevic: I would like to remove all those separate variables for classes before the html, they are hard to read and you need to jump up and down to change something. I'm ok with assigning this to a variable if it is something complicated but a s simple as this is I will like to have it in a inline form

blocks/init/src/blocks/components/button/components/button-options.js
iruzevic: So I will return it to how it was

blocks/init/src/blocks/components/paragraph/components/paragraph-options.js
IvanGrginovInf: that way you can render only relevant options when using paragraph inside another component.

Say you have a hero, using Header, Paragraph, Image & Button.

You only want the user to be able to change the Header / Paragraph's color but not size or anything else. Then you could just pass `onChangeStyleColor` to `paragraph-options.js` and only that option would be exposed to the user.

Hope that makes more sense.

blocks/init/src/blocks/wrapper/components/wrapper-editor.js
```diff
@@ -1,40 +1,41 @@
 import React from 'react'; // eslint-disable-line no-unused-vars
 
 export const WrapperEditor = (props) => {
```
gabriel-glo, line 3: These props can be destructured as Denis suggested.

blocks/init/src/blocks/custom/card-list/card-list.php
IvanGrginovInf: Fair enough but note we just did the exact oposite change in React components in this PR (moved inline code above `render` method) 😄

blocks/init/src/blocks/custom/video/video-block.js
```diff
@@ -4,6 +4,7 @@ import manifest from './manifest.json';
 import { VideoEditor } from './components/video-editor';
 
 export const Video = (props) => {
+
```
gabriel-glo, line 7: Destructure the props as Denis suggested

blocks/init/src/blocks/components/paragraph/components/paragraph-options.js
iruzevic: I have answered this in some PR before. But never mind. We are wrapping each option component inside a conditional because if you want for example to use paragraph option in you block and you only want to use Color and not the whole set of options you can just provide a prop for color and you have only colors. If this wan't in the conditions all other props will throw an error. So this is the way we have managed to make components as versatile as possible.

Why are we wrapping a condition with and onChange method. Because of a bool value. If you wrap a bool in a condition when it is set to false it will not show at all. So for consistency sake we have agreed to wrap everything inside a onChange method rather than value. 

In general inside a blockOption you don't need to have to have conditional around the options but in components you do. So also as a good practice an consistency we have added the conditionals inside a blocks also.

IvanGrginovInf:
> I have one question tho - why classnames package? Seems to be doing the same thing as was done by just listing classes 🤷‍♂️ I know that GB is using it, is that the only reason why or are there other benefits?

@dingo-d I'm sure there are better reasons but imo it provides a readable standardised way of concatenating multiple classes into a single string and in the end it flatens them. If you just concatenated multiple strings (classes) it would either be ugly code or they wouldn't be a single line in output html - depending on how you go about it.

blocks/init/src/blocks/components/button/components/button-options.js
iruzevic: > Left some comments, most are the same thing about destructuring props.
> 
> I have one question tho - why `classnames` package? Seems to be doing the same thing as was done by just listing classes 🤷‍♂️ I know that GB is using it, is that the only reason why or are there other benefits?

In general you don't need it but it gives you a much better control over the class names. You can write conditionals inside it so it looks nice. It is nice and separated.

blocks/init/src/blocks/custom/card-list/card-list.php
iruzevic: can you please give me an example? 
are you referring to the check for this:
```const buttonObject = (typeof button === 'undefined') || button;```

iruzevic:
I have returned all th js object destructuring to how it was before the PR as suggested from the JS team

blocks/init/src/blocks/custom/card-list/card-list.php
IvanGrginovInf: > `const buttonObject = (typeof button === 'undefined') || button;` 
yes, that was inline before (inside markup / render), we moved it above `render`.

In for example `card-list.php`, classes were defined above markup. Now they're moved inline. 

It's the exact oposite of the `const buttonObject = (typeof button === 'undefined') || button;` type changes.

IvanGrginovInf: I'm generally fine when it's simple (like here, only 2 classes) but we're just adding more inconsistencies between the way we write React and the way we write PHP components.

iruzevic: Yes but this is different. The changes for the undefined was done because you don't want a conditions inside a props that you are passing inside a component. As for this example this is only css selectors inline there are no conditions here. I there were some conditions I would move it out of the render methods just like in the PHP.

you can see here that Kuki also used classes inside the html inline:
https://infinum.slack.com/archives/C027S03F6/p1588695511204800

iruzevic: > I'm generally fine when it's simple (like here when it's only 2 classes) but we're just adding more inconsistencies between the way we write React and the way we write PHP components.

Can we agree that if the selectors became less readable inline you can extract them inside a variable before. If there are few like here where is 2 you can leave it inline?

IvanGrginovInf: > Can we agree that if the selectors became less readable inline you can extract them inside a variable before. If there are few like here where is 2 you can leave it inline?

Agreed.

blocks/init/src/blocks/components/paragraph/components/paragraph-options.js
dingo-d: Cool, makes sense. Thanks for the explanation!

PR #122, opened 2020-05-12
Feature/transforms

iruzevic:
* We now have ability to provide transforms a custom icons as svg inside block registration process.
* Added transforms for Heading and Paragraph Block

Custom Icon:
Screenshot 2020-05-12 at 15 05 19

Transforms:
[video.mov.zip](https://github.com/infinum/eightshift-frontend-libs/files/4615741/video.mov.zip)


scripts/editor/register-blocks.js
```diff
@@ -1,6 +1,60 @@
 import { registerBlockType } from '@wordpress/blocks';
 import { registerBlock } from './register-block';
-import { getBlockEditComponent } from './get-block-edit-component';
+
+/**
+ * Filter array of javascript paths and get the correct edit component.
```
gabriel-glo, line 5:
```suggestion
 * Filter array of JavaScript paths and get the correct edit component.
```

iruzevic: fixed

PR #123, opened 2020-05-12
adding option to have multiple mocked examples in storybook inside in…

iruzevic:
adding option to have multiple mocked examples in storybook inside innerblocks

### DIFF
--- a/blocks/init/src/blocks/custom/carousel/docs/story.js
+++ b/blocks/init/src/blocks/custom/carousel/docs/story.js
@@ -5,7 +5,7 @@ import React from 'react';
 import manifest from './../manifest.json';
 import globalManifest from './../../../manifest.json';
 import readme from './readme.md';
-import { block as item } from '../../carousel-image/docs/story';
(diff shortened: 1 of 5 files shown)


blocks/init/src/blocks/custom/carousel/docs/story.js
```diff
@@ -16,6 +16,6 @@ export default {
 };
 
 export const block = () => (
-  <Gutenberg props={blockDetails(manifest, globalManifest, item())} />
+  <Gutenberg props={blockDetails(manifest, globalManifest, [item(), item2(), item3()], 3)} />
```
kmaric69, line 19: what does the parameter 3 at the end mean? Is the number of items inside the array?

iruzevic: It is the number of times this items array will be duplicated.
This way you will have an output of

item1, item2, item3, item1, item2, item3, item1, item2, item3

PR #127, opened 2020-05-12
Feature/helpers

iruzevic:
* now to be able to have multiple onChange methods for object you had to write multipleProps: true key. Now this is no longer necessary

### DIFF
--- a/blocks/init/src/blocks/custom/button/manifest.json
+++ b/blocks/init/src/blocks/custom/button/manifest.json
@@ -34,8 +34,7 @@
       },
       "items": {
         "type": "object"
-      },
-      "multipleProps": true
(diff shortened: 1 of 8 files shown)


IvanGrginovInf:
> now to be able to have multiple onChange methods for object you had to write multipleProps: true key. Now this is no longer necessary

@iruzevic should these keys then be removed from all our blocks?

iruzevic:
> > now to be able to have multiple onChange methods for object you had to write multipleProps: true key. Now this is no longer necessary
> 
> @iruzevic should `multipleProps` key then be removed from all our block manifests?

In general yes, but if you leave it in it wont do anything so it is up to you

PR #129, opened 2020-05-12
adding changes to jumbotron and card-list for alight

iruzevic:
* Jumbotron
* Card-List

Screenshot 2020-05-13 at 00 47 10


dingo-d:
The 'delete image' button on the preview is overlapping the toolbar, this is a glitch or? I was thinking that maybe the delete image button would be better placed in the image toolbar?

gabriel-glo:
@dingo-d 
They are overlapping because this screenshot is from Storybook and there we don't have correct spacing for the block area. Adding some top padding would fix this.

iruzevic:
I have added remove button to toolbar in all component instances

Screenshot 2020-05-13 at 10 18 21

PR #131, opened 2020-05-13
Implement ability to toggle wrapper

iruzevic:
- Wrapper has option to disable its self from the start but only by code and not from editor. With this toggle you can for example add paragraph block inside a columns block and simply click on the toggle to remove wrapper and then this block behaves the same way as a normal block.

This is usefull inside innerBlocks

Screenshot 2020-05-14 at 00 12 49

### DIFF
--- a/blocks/init/src/blocks/wrapper/wrapper.js
+++ b/blocks/init/src/blocks/wrapper/wrapper.js
@@ -10,9 +10,6 @@ export const Wrapper = (props) => {
   const {
     props: {
       attributes,
-      attributes: {
-        hasWrapper,
-      },
     },
     children,
   } = props;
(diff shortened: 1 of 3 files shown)


blocks/init/src/blocks/wrapper/wrapper.js
```diff
@@ -10,9 +10,6 @@ export const Wrapper = (props) => {
   const {
     props: {
       attributes,
-      attributes: {
-        hasWrapper,
-      },
     },
     children,
   } = props;
```
gabriel-glo, line 15: Can these props be  destructured in function arguments?

iruzevic: not in a pretty an useful way

PR #164, opened 2020-07-13
Simplify getActions

IvanGrginovInf:
Now we no longer need to add `multipleProps: true / false` for `type: object` attributes. `getActions()` now returns both (whole object setters + individual property setters) and you can just pick and choose which ones you need.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ All notable changes to this project will be documented in this file.
 This projects adheres to [Semantic Versioning](https://semver.org/) and [Keep a CHANGELOG](https://keepachangelog.com/).
 
 ## [Unreleased]
+- Simplified building actions (`multipleProps: true|false` is no longer needed), `getActions()` now returns both multipleProp and singleProp actions (whole object setters + individual property setters) and you can just pick and choose which ones you need
+- Removed `mediaPropActions` which is redundant
 
 ## [3.5.0] - 2020-06-03
 
--- a/scripts/editor/get-actions.js
+++ b/scripts/editor/get-actions.js
@@ -119,7 +119,10 @@ export const getActions = (props, manifest) => {
           ...singlePropsAction(setAttributes, key),
         };
       } else {
-        actionsOutput = { ...actionsOutput, ...singlePropsAction(setAttributes, key) };
+        actionsOutput = {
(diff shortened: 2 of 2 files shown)


iruzevic:
Can you please write her some examples how this works?

IvanGrginovInf:
@iruzevic  Let's say you have an attribute in block's `manifest.json` like so:

```json
    "selectedPost": {
      "type": "object",
      "default": {
        "value": 0,
        "id": 0,
        "label": "No post selected"
      },
    },
```

Now you can get **both** individual property setter actions **as well as** the entire object setter action from `getActions()` like so (no need to write `multipleProps: true|false`):

```js
import { getActions } from '@eightshift/frontend-libs/scripts/editor';

const {
  onChangeSelectedPost,      // set the entire object
  onChangeSelectedPostId,    // set a single prop
  onChangeSelectedPostValue, // set a single prop
  onChangeSelectedPostLabel, // set a single prop
} = getActions(props, manifest);
```

dingo-d:
I can approve this but I'm not 100% sure how this works or what it does 😬 😅

IvanGrginovInf:
> I can approve this but I'm not 100% sure how this works or what it does 😬 😅

`getActions()` is a function which automatically builds actions (setters) for all block's attributes. All I've done is simplified the process by a tiny bit. 😄

iruzevic:
@IvanGrginovInf have you done that check where you only update the keys in the object that are changed and not the whole object. That way you can benefit from not setting the full object for defaulta?

IvanGrginovInf:
> have you done that check where you only update the keys in the object that are changed and not the whole object. That way you can benefit from not setting the full object for defaulta?

@iruzevic Yeah I tried implementing that but I don't believe this is possible at all. You need to set the entire object, you can't set just a single property. And by setting the entire object you must overwrite those defaults.

iruzevic:
> > have you done that check where you only update the keys in the object that are changed and not the whole object. That way you can benefit from not setting the full object for defaulta?
> 
> @iruzevic Yeah I tried implementing that but I don't believe this is possible at all. You need to set the entire object, you can't set just a single property. And by setting the entire object you must overwrite those defaults.

Ok, lets merge this for now and I will try in the next PR

PR #199, opened 2020-12-14
Fix path to the blocks folder for standalone Storybook

gabriel-glo:
The path to blocks folder was same as for the Storybook setup in boilerplate, which is different for standalone Storybook setup.


iruzevic (changes requested):
This PR is wrong. The original one was correct. I have tested it now and it is working. 

Please try cloning a new repo and check. This looks like some kind of case or cache issue

gabriel-glo:
It was the issue with cached version and git not realising that folder now has uppercased name.
I'll close this PR.

PR #201, opened 2020-12-14
Create _strip-unit.scss

dingo-d:
Add a strip unit helper that will be used for the calculation of the relative units. With this helper, we can pass numbers with units (e.g `20px`) or without, when calculating relative values (those will have to be changed of course). That way we can reuse things from `manifest.json`.

### DIFF
--- a/styles/scss/functions/_strip-unit.scss
+++ b/styles/scss/functions/_strip-unit.scss
@@ -1,0 +1,22 @@
+/// Strips the unit from the passed value
+///
+/// If context is not provided, it will be pulled from the $base-font-size.
+///
+/// ### Output
+/// ```scss
+/// $width: 960px;
+/// $unitless-width: strip-unit($width); // 960
+/// ```
+///
+/// @access public
+/// @author https://css-tricks.com/snippets/sass/strip-unit-function/
+/// @param {Number} $number The number to remove unit from.
+///
+
+@function strip-unit($number) {
+  @if type-of($number) == 'number' and not unitless($number) {
+    @return $number / ($number * 0 + 1);
+  }
+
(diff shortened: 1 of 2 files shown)


styles/scss/functions/_strip-unit.scss
```diff
@@ -0,0 +1,22 @@
+/// Strips the unit from the passed value
+///
+/// If context is not provided, it will be pulled from the $base-font-size.
```
IvanGrginovInf, line 3:
```suggestion
```

iruzevic (changes requested):
Greate you need to add this file to the functions/index.scss file as well

IvanGrginovInf (changes requested):

styles/scss/functions/_strip-unit.scss
dingo-d: This is still valid 👀

iruzevic (approved):

styles/scss/functions/_strip-unit.scss
IvanGrginovInf:
```scss
@function strip-unit($number) {
  @if type-of($number) == 'number' and not unitless($number) {
    @return $number / ($number * 0 + 1);
  }

  @return $number;
}
```

Where exactly is `context` used here?

IvanGrginovInf (changes requested):

dingo-d:
@IvanGrginovInf my bad, thought this was the em PR 🤦🏼‍♂️

IvanGrginovInf (approved):

iruzevic (approved):

PR #215, opened 2020-12-22
updating components

iruzevic:
I have updated the fix so now you can now override component from component from component and etc.
- a few fixes on blocks and components

### DIFF
--- a/scripts/editor/register-blocks.js
+++ b/scripts/editor/register-blocks.js
@@ -243,18 +243,18 @@ export const prepareComponentAttributes = (componentsManifest, blockManifest, bl
 
 			const realComponentName = components[newComponentName];
 
-			const component = componentsManifest.filter((item) => item.componentName === realComponentName);
+			const [component] = componentsManifest.filter((item) => item.componentName === realComponentName);
 
-			if (!component.length) {
+			if (!component) {
 				throw Error(`Component specified in "${blockName}" blocks manifest doesn't exist in your components list. Please check if you project has "${realComponentName}" component.`);
 			}
 
 			let outputAttributes = {};
 
-			if (component[0].hasOwnProperty('components')) {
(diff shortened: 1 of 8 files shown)


scripts/editor/register-blocks.js
```diff
@@ -237,14 +252,15 @@ export const prepareComponentAttributes = (componentsManifest, components, block
 			let outputAttributes = {};
 
 			if (component[0].hasOwnProperty('components')) {
-				outputAttributes = prepareComponentAttributes(componentsManifest, component[0].components, blockName, key);
+				outputAttributes = prepareComponentAttributes(componentsManifest, component[0], blockName, key);
```
volfkarlo, line 255: why not use destructuring here, something like 

```
const [component] = componentManifest.filter((item) => item.componentName === realComponentName);

if (component) {
throw Error(`Component specified in "${blockName}" blocks manifest doesn't exist in your components list. Please check if you project has "${realComponentName}" component.`);
			}
```
?
 By doing that, you can use component as a variable and no need to carry around the array of components.

iruzevic: this method is used to recursively run and in that method I need `components`, `attributes` and `example` so I can't only pass components key in the method

volfkarlo: yes, I understand. But I'm talking that line 246 can change from 

```
const component = componentsManifest.filter((item) => item.componentName === realComponentName);
```
to
```
const [component] = componentManifest.filter((item) => item.componentName === realComponentName);
```
because later you are using only, if I'm not mistaken, `component[0]`

iruzevic: a ha okej, yea this worked. I have updated it

kmaric69 (approved):

IvanGrginovInf (approved):

volfkarlo (approved):

PR #229, opened 2021-01-13
Feature/jumbotron

iruzevic:
- New block Jumbotron
- New component Jumbotron
- Changes on image align inside image-wrap

Screenshot 2021-01-13 at 15 49 51
de its wrapper

### DIFF
--- a/blocks/init/src/Blocks/components/image/image-style.scss
+++ b/blocks/init/src/Blocks/components/image/image-style.scss
@@ -15,7 +15,84 @@
 .image-wrap {
 	display: flex;
 
-	&__align {
-		@include flex-horizontal-align-modifiers;
+	&[data-align='top left'] {
+		align-items: flex-start;
+		justify-content: flex-start;
+
+		.image--bg {
+			background-position: left top;
+		}
+	}
+
+	&[data-align='top center'] {
+		align-items: flex-start;
+		justify-content: center;
+
+		.image--bg {
+			background-position: center top;
+		}
+	}
+
+	&[data-align='top right'] {
+		align-items: flex-start;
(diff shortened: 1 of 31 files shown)


blocks/init/src/Blocks/components/image/image-style.scss
```diff
@@ -15,7 +15,84 @@
 .image-wrap {
 	display: flex;
 
-	&__align {
-		@include flex-horizontal-align-modifiers;
+	&[data-align='top left'] {
```
dingo-d, line 18: This would be a good candidate for some kind of placeholder if we will use the matrix alignment in other components some day.

blocks/init/src/Blocks/components/jumbotron/docs/story.js
```diff
@@ -0,0 +1,80 @@
...
+
+export const options = () => (
+	<JumbotronOptions {...props} />
+);
+
+export const toolbar = () => (
+	<JumbotronToolbar {...props} />
+);
+
+const aligns = [
```
dingo-d, line 42: Seems like this is repeating. Would it be a good idea to move it to a helper so that it can be reused?

IvanGrginovInf (approved):

dingo-d (approved):
Left some comments, but all in all LGTM 👍🏼

blocks/init/src/Blocks/components/jumbotron/docs/story.js
iruzevic: Totally agree. I hope once this component is live the core will provide the helper but I agree we should make one until then

PR #230, opened 2021-01-13
Fix onChange property for videoType attribute.

gabriel-glo:
Fix typo in `videoType` attribute name
Add `help` property to the `videoType` select component
Remove `videoUrl` from `onChange` method of  `videoType` select component


iruzevic (approved):

dingo-d (approved):
On line 55 I've added `(videoType === 'local' && videoUrl === '') && videoUsePlaceholder ?` but I think it will work without it as well.

volfkarlo (approved):

PR #241, opened 2021-01-19
fix vertical spacing not being applied on Columns

goranalkovic-infinum:
_Vertical spacing_ was not being applied on the Columns block because of the wrong class name in the SCSS file.

### DIFF
--- a/blocks/init/src/Blocks/custom/columns/styles/_shared.scss
+++ b/blocks/init/src/Blocks/custom/columns/styles/_shared.scss
@@ -30,7 +30,7 @@ $block-columns: (
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


blocks/init/src/Blocks/custom/columns/styles/_shared.scss
```diff
@@ -30,7 +30,7 @@ $block-columns: (
 		}
 	}
 
-	&__vertical-spacing {
+	&__verticalSpacing {
```
dingo-d, line 33: To me this looks like a problem in how the class name is generated in. We don't write SCSS with camel case 👀

PR #246, opened 2021-01-26
updating column with align

iruzevic:
- Removed duplicated calls for scripts
- Added option on column to vertically align so you can get this kind of design
Screenshot 2021-01-26 at 17 46 57

### DIFF
--- a/blocks/init/src/Blocks/custom/column/components/column-options.js
+++ b/blocks/init/src/Blocks/custom/column/components/column-options.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import { __ } from '@wordpress/i18n';
 import { Fragment } from '@wordpress/element';
-import { PanelBody, Icon, RangeControl, ToggleControl } from '@wordpress/components';
+import { PanelBody, Icon, RangeControl, ToggleControl, SelectControl } from '@wordpress/components';
 import { icons, ucfirst } from '@eightshift/frontend-libs/scripts/editor';
 import { Responsive, HelpModal } from '@eightshift/frontend-libs/scripts/components';
 import manifest from './../manifest.json';
@@ -28,6 +28,13 @@ export const ColumnOptions = ({ attributes, setAttributes }) => {
 		attributes.orderDesktop,
 		attributes.orderTablet,
 		attributes.orderMobile,
+	];
+
+	const align = [
+		attributes.alignLarge,
+		attributes.alignDesktop,
+		attributes.alignTablet,
+		attributes.alignMobile,
 	];
 
 	const hide = [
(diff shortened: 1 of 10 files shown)


blocks/init/src/Blocks/custom/column/manifest.json
```diff
@@ -89,6 +89,19 @@
 		},
 		"orderMobile": {
 			"type": "integer"
+		},
+
```
dingo-d, line 93: Why the extra space?

dingo-d (changes requested):
One small change (empty space in the manifest file)

gabriel-glo (approved):

blocks/init/src/Blocks/custom/column/manifest.json
```diff
@@ -112,6 +125,20 @@
 			"min":  1,
 			"max":  20,
 			"step": 1
-		}
+		},
+		"aligns": [
```
goranalkovic-infinum, line 129: Would it be a good idea to include the usual flexbox default (stretch)?

blocks/init/src/Blocks/custom/column/components/column-options.js
```diff
@@ -138,6 +145,32 @@ export const ColumnOptions = ({ attributes, setAttributes }) => {
...
+				}
+			>
+				{align.map((item, index) => {
+
+					const point = ucfirst(options.breakpoints[index]);
+					const attr = `align${point}`;
+
+					return (
+						<Fragment key={index}>
+							<SelectControl
```
goranalkovic-infinum, line 163: Would a `ButtonGroup` with `IconButton`s inside have better UX than a `SelectControl`?

e.g.:
image

blocks/init/src/Blocks/custom/column/manifest.json
iruzevic: All item are grouped for responsive so it is easier to preview and manage

blocks/init/src/Blocks/custom/column/components/column-options.js
iruzevic: Sure it would. Make an PR for this change

blocks/init/src/Blocks/custom/column/manifest.json
iruzevic: Check this whole file

dingo-d: Didn't notice the rest, GH collapses things. 👍🏼

dingo-d (approved):
LGTM, I like the suggestions Goran made 🙂

blocks/init/src/Blocks/custom/column/components/column-options.js
iruzevic: i like it

PR #247, opened 2021-01-28
Add icon component

kmaric69:
I've used the icon component that @JerrySpecter created and added it to our frontend libs.


dingo-d (dismissed):

blocks/init/src/Blocks/components/icon/manifest.json
```diff
@@ -0,0 +1,113 @@
...
...52 12 21.7726C11.787 21.6152 11.5211 21.4133 11.2167 21.1708C10.3987 20.5195 9.31061 19.5822 8.22595 18.4373C6.01574 16.1042 4 13.0981 4 10ZM11.4448 23.8317C11.445 23.8319 11.4453 23.8321 12 23L11.4453 23.8321C11.7812 24.056 12.2188 24.056 12.5547 23.8321L12 23C12.5547 23.8321 12.555 23.8319 12.5552 23.8317L12.556 23.8312L12.5581 23.8298L12.5648 23.8253L12.5877 23.8098C12.6072 23.7966 12.6349 23.7776 12.6704 23.753C12.7415 23.7038 12.8435 23.6321 12.9722 23.5392C13.2295 23.3534 13.5936 23.0822 14.0292 22.7354C14.8987 22.043 16.0606 21.0428 17.226 19.8127C19.5157 17.3958 22 13.9019 22 10C22 4.47715 17.5228 0 12 0C6.47715 0 2 4.47715 2 10C2 13.9019 4.48426 17.3958 6.77405 19.8127C7.93939 21.0428 9.10133 22.043 9.97082 22.7354C10.4064 23.0822 10.7705 23.3534 11.0278 23.5392C11.1565 23.6321 11.2585 23.7038 11.3296 23.753C11.3651 23.7776 11.3928 23.7966 11.4123 23.8098L11.4352 23.8253L11.4419 23.8298L11.444 23.8312L11.4448 23.8317ZM12 6C9.79086 6 8 7.79086 8 10C8 12.2091 9.79086 14 12 14C14.2091 14 16 12.2091 16 10C16 7.79086 14.2091 6 12 6ZM10 10C10 8.89543 10.8954 8 12 8C13.1046 8 14 8.89543 14 10C14 11.1046 13.1046 12 12 12C10.8954 12 10 11.1046 10 10Z' fill='currentColor' /></svg>",
```
goranalkovic-infinum, line 97: Do we need to have the `-icon` suffix?  Seems a bit redundant.

IvanGrginovInf (approved):

blocks/init/src/Blocks/components/icon/components/icon-options.js
```diff
@@ -0,0 +1,74 @@
...
+import { SelectControl, ToggleControl, Icon } from '@wordpress/components';
+import { ColorPaletteCustom } from '@eightshift/frontend-libs/scripts/components';
+import { checkAttr } from '@eightshift/frontend-libs/scripts/helpers';
+import { getPaletteColors, icons } from '@eightshift/frontend-libs/scripts/editor';
+import manifest from './../manifest.json';
+
+const { options, title } = manifest;
+
+export const iconColors = () => {
+	const colors = getPaletteColors();
```
iruzevic, line 13: use new helper for this. Check button for inspiraton

blocks/init/src/Blocks/components/icon/components/icon-options.js
```diff
@@ -0,0 +1,74 @@
...
+						options={options.icons}
+						onChange={(value) => setAttributes({ [`${componentName}Name`]: value })}
+					/>
+
+					<ColorPaletteCustom
+						label={
+							<Fragment>
+								<Icon icon={icons.color} />
+								{__('Icon color', 'eightshift-frontend-libs')}
+
```
iruzevic, line 62: uncecesery space

blocks/init/src/Blocks/components/icon/docs/story.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react'; // eslint-disable-line no-unused-vars
+import { Fragment } from '@wordpress/element';
+import readme from './readme.md';
+import manifest from './../manifest.json';
+import { IconEditor } from '../components/icon-editor';
+import { IconOptions } from '../components/icon-options';
+
+export default {
+	title: `Components/${manifest.title}`,
+	parameters: {
```
iruzevic, line 10: this has changed in the lates release. please check other components for fix

blocks/init/src/Blocks/components/icon/icon.php
```diff
@@ -0,0 +1,41 @@
...
+ * @package EightshiftBoilerplate
+ */
+
+use EightshiftBoilerplateVendor\EightshiftLibs\Helpers\Components;
+
+$manifest = Components::getManifest(__DIR__);
+$componentName = $attributes['componentName'] ?? $manifest['componentName'];
+
+$iconUse = Components::checkAttr('iconUse', $attributes, $manifest, $componentName);
+
```
iruzevic, line 15: remove space

blocks/init/src/Blocks/components/icon/manifest.json
```diff
@@ -0,0 +1,113 @@
...
+			"type": "string",
+			"default": "primary"
+		},
+		"iconName": {
+			"type": "string",
+			"default": "play-icon"
+		},
+		"iconUse": {
+			"type": "boolean",
+			"default": false
```
iruzevic, line 23: why is this false?

blocks/init/src/Blocks/components/icon/manifest.json
iruzevic: agreed

iruzevic (changes requested):

gabriel-glo (approved):

dingo-d (approved):

blocks/init/src/Blocks/components/icon/components/icon-options.js
```diff
@@ -0,0 +1,62 @@
...
+	if (!showIconOptions) {
+		return null;
+	}
+
+	return (
+		<Fragment>
+			{label && <h3 className={'options-label'}>{label}</h3>}
+
+			<ToggleControl
+				label={__('Use Icon', 'eightshift-frontend-libs')}
```
goranalkovic-infinum, line 32: This shouldn't be hardcoded.

Write it like this:
`label={sprintf(__('Use %s', 'eightshift-frontend-libs'), label)}`

blocks/init/src/Blocks/components/icon/icon-style.scss
```diff
@@ -0,0 +1,45 @@
+$icon: (
+	colors: (
+		primary: (
+			normal: (
+				color: global-settings(colors, primary),
+			),
+			hover: (
```
goranalkovic-infinum, line 7: Do all icons need to have a hover effect by default?

They're all set to `currentColor` for fills/strokes, so for most cases regular color option is more than enough.
For cases when hover style change is needed the icon's color can be easily overridden.

goranalkovic-infinum (changes requested):

blocks/init/src/Blocks/components/icon/docs/story.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react'; // eslint-disable-line no-unused-vars
+import { Fragment } from '@wordpress/element';
+import readme from './readme.md';
```
iruzevic, line 3: you must include mdx file. How is this working in Storybook? have you tested it?

iruzevic (changes requested):

blocks/init/src/Blocks/components/icon/components/icon-options.js
kmaric69: Good catch

blocks/init/src/Blocks/components/icon/docs/story.js
kmaric69: Fixed the include and tested in storybook. All seems good now.

blocks/init/src/Blocks/components/icon/icon-style.scss
kmaric69: Makes sense..no need for all color variations. I changed it by leaving only primary and black color.

goranalkovic-infinum: Sorry, looks like there was a misunderstanding. The multiple color variants were fine (even allowing all manifest colors would be fine).

I was asking about the need for a (default) hover state for the icon. In most cases the icon will be either a decoration by itself or a part of something interactive (like a button).

If the icon is a part of something else (e.g. a Button), because the SVGs for all the icons have `currentColor` as their fill/stroke colors, they would pick up the surrounding text color (can also be set as `color: inherit` in CSS).
By inheriting the text color, you don't need to worry about changing the hover/focus/disabled/etc. color because that would be handled by the parent component.

In cases where the icon should be decorative or by itself (e.g. a Card or List), a color option is useful, but in those cases the hover color probably wouldn't be needed.

dingo-d: Good point. This was taken from my implementation because I had a case where only an icon was used so I needed a hover. I'll have to change this implementation, as the hover looks odd when used in a compound setting (as a part of another block). 👍🏼

iruzevic (dismissed):

IvanGrginovInf (commented):
Gotta say now that I look / think about it, I'm not really a fan of this. For example looking at `blocks/init/src/Blocks/components/icon/manifest.json`, we need to load all icons as soon as you use this component.

This is starting to remind me of Elementor where when you include even a single icon from a pack in your project, it will load the entire CSS for the entire pack. So there you add 1 icon and immediately get 200+ kb of CSS.

Maybe I'm crazy, not even sure if this is going to be a problem but pretty sure the way it's written currently (unless you modify the component) you will need to load all icons inside even if you just need one. I mean it's just in the editor but React will still need to load all this `svg` markup from `manifest.json` on each page reload.

Thoughts? Is this even an issue or am I being paranoid?

dingo-d:
We can profile the calls to see if it has a noticeable effect on the page speeds. The icons are loaded as text (SVG string) inside a manifest, which is read and then used from there. So the only thing that loads is `manifest.json`, and I think that GB has a smart way of loading this once (I sure hope it does 😂 ). 

The way we could test this is by seeing how the size of the manifest changes as we load more and more icons. For instance, what's the current size, then just duplicate the same icons 100 times and see the file size difference and then try to see how this impacts the editor and the frontend.

blocks/init/src/Blocks/components/icon/icon-style.scss
kmaric69: Ok, would it be fine with you that I remove the color options from the component and just set `color: inherit` inside the css?

kmaric69: Because I think the component will mostly be used as part of a parent component.

dingo-d (approved):

blocks/init/src/Blocks/components/icon/icon-style.scss
goranalkovic-infinum: Good idea!

gabriel-glo (dismissed):

dingo-d (approved):

goranalkovic-infinum (approved):

gabriel-glo (approved):

PR #248, opened 2021-01-28
Cleanup

dingo-d:
I've done the cleanup inspection for the scripts part of the frontend libs.

There are lots of minor fixes. The easiest way to look at this PR is to go through the commits when reviewing to see what has been changed, and to understand the reason for the fixes.

Screenshot 2021-01-29 at 08 50 31

### DIFF
--- a/blocks/init/src/Blocks/components/accordion/assets/accordion.js
+++ b/blocks/init/src/Blocks/components/accordion/assets/accordion.js
@@ -12,7 +12,6 @@ export class Accordion {
 
 		this.ATTR_HIDDEN = 'aria-hidden';
 		this.ATTR_OPEN = 'data-accordion-open';
-		this.ATTR_PREVENT_CLOSE = 'data-accordion-prevent-close';
 	}
 
 	init() {
--- a/blocks/init/src/Blocks/components/image/components/image-editor.js
+++ b/blocks/init/src/Blocks/components/image/components/image-editor.js
@@ -19,6 +19,7 @@ export const ImageEditor = (attributes) => {
 		imageUse = checkAttr('imageUse', attributes, manifest, componentName),
 
 		imageUrl = checkAttr('imageUrl', attributes, manifest, componentName),
+		imageAlt = checkAttr('imageAlt', attributes, manifest, componentName),
 		imageAccept = checkAttr('imageAccept', attributes, manifest, componentName),
 		imageAllowedTypes = checkAttr('imageAllowedTypes', attributes, manifest, componentName),
 		imageBg = checkAttr('imageBg', attributes, manifest, componentName),
@@ -46,7 +47,7 @@ export const ImageEditor = (attributes) => {
 							<Fragment>
 								{imageBg ?
 									<div className={imageClass} style={{ backgroundImage: `url(${imageUrl})` }} /> :
-									<img className={imageClass} src={imageUrl} alt="" />
+									<img className={imageClass} src={imageUrl} alt={imageAlt} />
 								}
 							</Fragment>
 						}
@@ -56,7 +57,12 @@ export const ImageEditor = (attributes) => {
 								{(!imageUsePlaceholder) ?
 									<MediaPlaceholder
 										icon="format-image"
-										onSelect={(value) => setAttributes({ [`${componentName}Url`]: value.url })}
+										onSelect={(value) => {
+											setAttributes({
+												[`${componentName}Url`]: value.url,
+												[`${componentName}Alt`]: value.alt
+											});
+										}}
 										accept={imageAccept}
 										allowedTypes={imageAllowedTypes}
 									/> :
--- a/blocks/init/src/Blocks/components/image/components/image-options.js
+++ b/blocks/init/src/Blocks/components/image/components/image-options.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import { Fragment } from '@wordpress/element';
 import { __, sprintf } from '@wordpress/i18n';
-import { ToggleControl, Icon } from '@wordpress/components';
+import { ToggleControl, Icon, TextControl } from '@wordpress/components';
 import { MediaPlaceholder, URLInput } from '@wordpress/block-editor';
 import { icons } from '@eightshift/frontend-libs/scripts/editor';
 import { checkAttr } from '@eightshift/frontend-libs/scripts/helpers';
@@ -19,6 +19,7 @@ export const ImageOptions = (attributes) => {
 		imageUse = checkAttr('imageUse', attributes, manifest, componentName),
 
 		imageUrl = checkAttr('imageUrl', attributes, manifest, componentName),
+		imageAlt = checkAttr('imageAlt', attributes, manifest, componentName),
 		imageLink = checkAttr('imageLink', attributes, manifest, componentName),
 		imageAccept = checkAttr('imageAccept', attributes, manifest, componentName),
 		imageAllowedTypes = checkAttr('imageAllowedTypes', attributes, manifest, componentName),
@@ -26,6 +27,7 @@ export const ImageOptions = (attributes) => {
 		imageUsePlaceholder = checkAttr('imageUsePlaceholder', attributes, manifest, componentName),
 
 		showImageUrl = true,
+		showImageAlt = true,
 		showImageLink = true,
 		showImageBg = true,
 	} = attributes;
@@ -55,7 +57,10 @@ export const ImageOptions = (attributes) => {
 						<MediaPlaceholder
 							icon="format-image"
 							onSelect={(value) => {
-								setAttributes({ [`${componentName}Url`]: value.url });
+								setAttributes({
+									[`${componentName}Url`]: value.url,
+									[`${componentName}Alt`]: value.alt
+								});
 							}}
 							accept={imageAccept}
 							allowedTypes={imageAllowedTypes}
@@ -64,9 +69,17 @@ export const ImageOptions = (attributes) => {
 
 					<br />
 
+					{showImageAlt &&
+						<TextControl
+							label={__('Image alt tag', 'eightshift-frontend-libs')}
+							value={imageAlt}
+							onChange={(value) => setAttributes({ [`${componentName}Alt`]: value })}
+						/>
+					}
+
 					{showImageBg &&
 						<ToggleControl
-							label={__('Use as a background Image', 'eightshift-frontend-libs')}
+							label={__('Use as a background image', 'eightshift-frontend-libs')}
 							checked={imageBg}
 							onChange={(value) => setAttributes({ [`${componentName}Bg`]: value })}
 						/>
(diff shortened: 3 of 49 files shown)


iruzevic (approved):

goranalkovic-infinum (approved):

iruzevic (approved):

scripts/helpers/escape-string.js
```diff
@@ -1,5 +1,3 @@
-export const escapeString = {
-	escapeString(string) {
-		return string.replace(/([;&,.+*~':"!^#$%@[\]()=>|])/g, '\\$1');
-	},
-};
+export function	escapeString(string) {
```
kmaric69: some weard spacing here...should this be like this?

scripts/helpers/cookies.js
```diff
@@ -10,6 +22,14 @@ export const cookies = {
 
 		document.cookie = `${key}=${value};${pathValue}expires=${expires.toUTCString()}`;
 	},
+
+	/**
+	 * get a cookie
```
kmaric69, line 27: Get a cookie

kmaric69:
Left some minor comments. Rest looks good...nice job!

kmaric69 (dismissed):

scripts/helpers/escape-string.js
dingo-d: I think I've changed this 🙂

iruzevic (approved):

gabriel-glo (approved):

blocks/init/src/Blocks/components/image/manifest.json
```diff
@@ -37,6 +38,9 @@
 			"type": "boolean",
 			"default": false
 		},
+		"imageAlt": {
```
IvanGrginovInf, line 41: If added, this should probably be set in the block as well (taken from the image's alt tag and overridable) but probably out of scope for this PR.

blocks/init/src/Blocks/components/search-bar/search-bar.php
```diff
@@ -38,10 +42,14 @@
 	class="<?php echo \esc_attr($searchClass); ?>"
 	action="<?php echo \esc_url($searchBarAction); ?>"
 >
+	<?php if ($searchBarLabelShow) { ?>
+		<label for="<?php echo \esc_attr($searchBarId); ?>"><?php echo \esc_html($searchBarLabel); ?></label>
+	<?php } ?>
```
IvanGrginovInf, line 47: Arguably `<label>` should be set even if it's not shown (accessibility, screen reader should know what this input is for). But not sure if we need to support this functionality out of the box in the component or does this fall more in the line of "You need to set this properly in your project if you're not showing the label" 🤷‍♂️

blocks/init/src/Blocks/components/video/video.php
```diff
@@ -65,7 +65,6 @@
 		<iframe
 		class="<?php echo \esc_attr($videoClass); ?>"
 		src="<?php echo \esc_url($localUrl); ?>"
-		frameBorder="0"
```
IvanGrginovInf, line 68: Isn't this some compatibility thing / reset for older browsers (IE11?)

blocks/init/src/Blocks/components/share/assets/share.js
```diff
@@ -5,26 +5,26 @@ export class Share {
 
 	init() {
 		[...this.shareLinks].forEach((shareLink) => {
-			shareLink.addEventListener('click', (event) => {
+			shareLink.addEventListener('click', async (event) => {
 				event.preventDefault();
 
-				this.openShareDialog(event.currentTarget);
+				await this.openShareDialog(event.currentTarget);
```
IvanGrginovInf, line 11: It's a little weird that opening a modal should be async but ok 😄  (I assume because `navigator.share ` runs asyncronously?)

blocks/init/src/Blocks/components/accordion/assets/accordion.js
```diff
@@ -12,7 +12,6 @@ export class Accordion {
 
 		this.ATTR_HIDDEN = 'aria-hidden';
 		this.ATTR_OPEN = 'data-accordion-open';
-		this.ATTR_PREVENT_CLOSE = 'data-accordion-prevent-close';
```
IvanGrginovInf, line 15: Sure this is unsued?

IvanGrginovInf (approved):
Left a few comments, otherwise looks good

blocks/init/src/Blocks/components/image/manifest.json
dingo-d: Not sure will check it out. Block should be the place where alt tag is set (passed to the component)

blocks/init/src/Blocks/components/search-bar/search-bar.php
dingo-d: Definitely should be set out of the box so that we have a11y compliance covered.

By your logic, everything could be set in your project, but if we are exposing certain things in our libs, it's our responsibility to make it accessible, no?

blocks/init/src/Blocks/components/video/video.php
dingo-d: phpstorm said it's deprecated, and I don't think we should bother with IE11 support (even MS killed it 🤷🏼‍♂️ )

blocks/init/src/Blocks/components/share/assets/share.js
dingo-d: Correct, I got an error because of the navigator API (which is super experimental from what I've read) 😬

blocks/init/src/Blocks/components/accordion/assets/accordion.js
dingo-d: yup, couldn't find an instance of this anywhere 🤷🏼‍♂️

blocks/init/src/Blocks/components/search-bar/search-bar.php
IvanGrginovInf: >  it's our responsibility to make it accessible, no?

Yeah I agree for sure. Tho then we need to change this so that `<label>` is always rendered but if `$searchBarLabel === false` we just give it a class that hides the visually hides text. (https://infinum.github.io/eightshift-frontend-libs/sassdocs/#placeholder-visually-hidden - `%visually-hidden`)

IvanGrginovInf: Also then the markup needs to be something like (so we can **visually** hide `.label-text` if needed):

```html
<label>
  <span class="label-text">Some label</span>
  <input />
</label>
```

dingo-d: Good suggestion, I'll change this

IvanGrginovInf (approved):

gabriel-glo (dismissed):

blocks/init/src/Blocks/components/search-bar/search-bar-style.scss
```diff
@@ -5,12 +5,19 @@
...
 
+	&__label {
+		margin-right: 15px;
+
+		&--hidden {
+			  @extend %visually-hidden;
+		}
+	}
+
 	&__input {
```
goranalkovic-infinum, line 16: This should be in a map, not directly applied.

goranalkovic-infinum (changes requested):
Besides the `search-bar-style` change, everything else seems fine

blocks/init/src/Blocks/components/search-bar/search-bar-style.scss
dingo-d: Extending a placeholder? 👀

blocks/init/src/Blocks/components/image/components/image-editor.js
```diff
@@ -56,7 +57,12 @@ export const ImageEditor = (attributes) => {
 								{(!imageUsePlaceholder) ?
 									<MediaPlaceholder
 										icon="format-image"
-										onSelect={(value) => setAttributes({ [`${componentName}Url`]: value.url })}
+										onSelect={(value) => {
+											setAttributes({
+												[`${componentName}Url`]: value.url,
+												[`${componentName}Alt`]: value.alt
```
iruzevic, line 63: we should not store alt tag in the DB when saving the image. Because if you change alt in the media this will not change in any of the pages.

In the admin alt tag is not important only on frontend. If you need to fetch alt you can use this function https://developer.wordpress.org/reference/functions/attachment_url_to_postid/ to fetch post ID and attachment meta on the frontend.

iruzevic: imageALT is ok to provide as a pro to a component but not to store it in the db

blocks/init/src/Blocks/components/search-bar/search-bar-style.scss
iruzevic: map is not mandatory to use so this is ok

iruzevic (changes requested):

blocks/init/src/Blocks/components/image/components/image-editor.js
dingo-d: The alt tag is stored in the database when you edit the image in the media library, as a post meta 👀 . So if you add it there, it will be pulled from the image automatically. And it's necessary to have it so that the screen readers can interpret the image (a11y).

And @JerrySpecter noticed that the `attachment_url_to_postid` function is not working all the time for some reason 🤷🏼‍♂️

iruzevic: The way you did it is stored in the post_content field in the DB no the post_meta. This way on the frontend you will never get the updated alt if changed in the media

iruzevic (changes requested):

goranalkovic-infinum (dismissed):

kmaric69 (dismissed):

dingo-d:
Hmmm, so I need to find a way to update the meta field for the alt when changed in the options 🤔  on my way!

blocks/init/src/Blocks/components/image/components/image-options.js
```diff
@@ -55,7 +57,10 @@ export const ImageOptions = (attributes) => {
 						<MediaPlaceholder
 							icon="format-image"
 							onSelect={(value) => {
-								setAttributes({ [`${componentName}Url`]: value.url });
+								setAttributes({
+									[`${componentName}Url`]: value.url,
+									[`${componentName}Alt`]: value.alt
```
iruzevic, line 62: you are setting alt in the TextareaControl so you dont need to do it here

blocks/init/src/Blocks/components/image/components/image-editor.js
iruzevic, line 63: alt is set in the options so no need to set it here

iruzevic (changes requested):
Also in the image-toolbar.js you once removing the image should we remove the alt as well?

blocks/init/src/Blocks/components/image/components/image-editor.js
dingo-d: This actually has to be set, otherwise the alt won't be shown. Not 100% sure why. I've removed it from the options and it works, but when I remove it from the editor, the alt text is not displayed in the editor or the options textarea 🤷🏼‍♂️

iruzevic (approved):

goranalkovic-infinum (approved):

PR #255, opened 2021-02-01
Column vertical align UX improvements

goranalkovic-infinum:
Adds toggle buttons instead of the select menu that is there currently.
image

Separate options are available for responsive sizes:
image

The _default_ option is `stretch` (when the value is `undefined`), which removes the class and leaves the default behaviour, enabling to user to opt out of any align options (which was not previously possible)


iruzevic:
Looks good but you have a lot of issue with gitAction check please fix them first

goranalkovic-infinum:
> Looks good but you have a lot of issue with gitAction check please fix them first

Fixed!
Gotta check my linter config 😄

dingo-d (approved):

kmaric69 (approved):

gabriel-glo (approved):

PR #267, opened 2021-02-03
updating browserSync https

iruzevic:
I have updated the Webpack so you can now in webpack.config.js provide only useSsl: true key and it will automatically convert browserSync to https. No need for custom setup


gabriel-glo (approved):

goranalkovic-infinum (approved):

dingo-d (approved):
Noice!

PR #269, opened 2021-02-04
Better tests setup

IvanGrginovInf:
This PR does the following:
- Makes Jest use our `babel/babel.config.js` so it can work with `import`'s
- Adds `__mocks__` folder for mocking `node_modules` (https://jestjs.io/docs/en/manual-mocks#mocking-node-modules)
- Added a mock for `import { useSelect } from @wordpress/data;` and all methods / properties needed for `optionColors` test: `select('core/block-editor').getSettings().colors` 
- Makes tests great again

What's cool about this is that we can now easily mock native Gutenberg methods / hooks (such as `useSelect()`) and return the value as we need for tests (this means we don't need to mock all functions which use hooks! 🎉 )


iruzevic (approved):

dingo-d (approved):
🎉

PR #275, opened 2021-02-09
Fix css class and selector

kmaric69:
The anchor smooth scroll didn't work properly because the class was wrong set inside the `button.php` file and the selector was wrong inside the `index.js` file


iruzevic (commented):
Can you check if this is also an issue on Link component?

dingo-d (approved):

iruzevic (approved):

IvanGrginovInf (approved):

PR #276, opened 2021-02-09
Paragraph UX improvements

### LINKED ISSUE #263: [FEATURE] Figure out how to correctly paste the content
In core WP, when you select content, either from a document or a website, and paste it, it's correctly parsed and placed in corresponding blocks.

This is a lifesaver.

We, unfortunately, don't have this ability out of the box. When I paste something I can only paste it directly in a block (paragraph), and then it's all one paragraph block, instead of multiple blocks for every paragraph.

Also, we need to find a way to add a default block as selected and add it automatically so that we can utilize the `/` command like in the core to call on the blocks.

This is a huge improvement on the a11y and overal UX.

goranalkovic-infinum:
Closes #263.

Eightshift Paragraph block now behaves way more like the vanilla Gutenberg one!

# Paragraph improvements

## Block splitting/merging

Now when you press _Enter_ a new block is created below, and if there was text in front of the cursor that text moves to the new block. Similarly, if you backspace (or forward-delete) between blocks, they get merged together.

## Proper copy-paste

As blocks are properly merged/split now, a custom paste handler was added that can process either plain text or HTML and paste it into the paragraph, keeping some formats¹ (like bold, italic, ...). On new lines (or paragraphs) in the pasted content, new blocks are created, just like when manually pressing Enter.

In the **component** manifest, you can optionally define which tags are allowed after parsing the pasted text:
```json
	"pasteAllowTags": [
		"strong",
		"em",
		"b",
		"i",
		"span"
	]
```
The defaults are shown above.

# Modular merging system

As Eightshift uses dynamic blocks, and merging requires a custom function on block registration, a new system was implemented that allows any² block to have merging enabled.

In the **block** manifest you can add a new key:
```json
	"mergeableAttributes": [
		{
			"attribute": "paragraphContent",
			"mergeStrategy": "append"
		}
	]
```
where you can define which properties are merged and how.

For now 6 merge strategies are implemented:
- `useSourceAttribute` (default) - takes the attribute value from the source object
- `useDestinationAttribute` - takes the attribute value from the destination object
- `append` - appends the destination's value to the source's value
- `addNumericIntValue` - adds source's and destination's values (parses to `int`) - ⚠️ make sure the value can be parsed!
- `addNumericFloatValue` - adds source's and destination's values (parses to `float`) - ⚠️ make sure the value can be parsed!
- `addNumericPixelValue` - takes only numbers (`int`) from the value, adds them and suffixes it with `px`

---
¹ As defined in `pasteAllowTags` (or default)
² Any block that uses `RichtText` in the editor and is sensible to be merged. Currently limited to same-type blocks!


blocks/init/src/Blocks/components/paragraph/components/paragraph-editor.js
```diff
@@ -12,8 +14,12 @@ export const ParagraphEditor = (attributes) => {
 		componentName = manifest.componentName,
 		componentClass = manifest.componentClass,
 		selectorClass = componentClass,
+		onSplit,
+		mergeBlocks,
+		onReplace,
+		onRemove,
 		blockClass,
-		placeholder = __('Add Content', 'eightshift-frontend-libs'),
+		placeholder = __('Add Content', 'Tema'),
```
IvanGrginovInf, line 22: @goranalkovic-infinum wrong namespace?

blocks/init/src/Blocks/components/paragraph/components/paragraph-editor.js
```diff
@@ -31,15 +39,152 @@ export const ParagraphEditor = (attributes) => {
 		selector(blockClass, blockClass, selectorClass),
 	]);
 
+	/**
+	 * Returns string parsed as nodes, ready for inserting into RichText.
+	 *
+	 * @param {string} inputText HTML string for parsing.
+	 */
+	const processTag = (inputText) => {
```
IvanGrginovInf, line 47: Could you maybe move all of these functions to separate files in the `blocks/init/src/Blocks/components/paragraph/components` folder and then import it from those files in `paragraph-editor`? That way we can keep the `paragraph-editor` more clean.

IvanGrginovInf (dismissed):
Left a comment regarding keeping `paragraph-editor` as clean as possible. Otherwise good job, this is awesome!

blocks/init/src/Blocks/components/paragraph/components/paragraph-editor.js
```diff
@@ -31,15 +39,152 @@ export const ParagraphEditor = (attributes) => {
...
+
+		return output;
+	}
+
+	/**
+	 * Process content and paste in into blocks smartly.
+	 *
+	 * @param {string} textFromClipboard Text retrieved from clipboard.
+	 */
+	const handlePaste = async (textFromClipboard) => {
```
dingo-d, line 80: You've tagged this as `async` but I don't see `await` anywhere. What am I missing?

blocks/init/src/Blocks/components/paragraph/components/paragraph-editor.js
```diff
@@ -31,15 +39,152 @@ export const ParagraphEditor = (attributes) => {
...
+	 *
+	 * @param {Event} event Passed event parameters.
+	 */
+	const pasteInto = (event) => {
+		event.preventDefault();
+		const [copiedData, copiedHtml] = event.clipboardData.items;
+
+		// If HTML content is in clipboard, process that...
+		if (copiedHtml) {
+			copiedHtml.getAsString(async (inputText) => {
```
dingo-d, line 139: Again, not seeing the `await` 🤷🏼‍♂️

blocks/init/src/Blocks/components/paragraph/components/paragraph-editor.js
```diff
@@ -31,15 +39,152 @@ export const ParagraphEditor = (attributes) => {
...
+
+					if (allowedTags.includes(filteredTagName)) {
+						inputText = inputText.replace(tag[0], parsedTag).replace(' >', '>');
+					}
+					else if ((filteredTagName === 'p' && !tag[0].includes('/')) || filteredTagName === 'br') {
+						inputText = inputText.replace(tag[0], '\n');
+					}
+					else {
+						inputText = inputText.replace(tag[0], '');
+					}
```
dingo-d, line 158:
```suggestion
					if (allowedTags.includes(filteredTagName)) {
						inputText = inputText.replace(tag[0], parsedTag).replace(' >', '>');
					} else if ((filteredTagName === 'p' && !tag[0].includes('/')) || filteredTagName === 'br') {
						inputText = inputText.replace(tag[0], '\n');
					} else {
						inputText = inputText.replace(tag[0], '');
					}
```

Just formatted to be a bit prettier.

dingo-d (changes requested):
Left some questions inline.

Native GB can handle paste of markdown (markdown -> blocks). Will this do the same? 👀

blocks/init/src/Blocks/components/paragraph/components/paragraph-editor.js
goranalkovic-infinum: You're not missing anything, I needed it at one point, but should've checked at the end if I still need it 😅

goranalkovic-infinum:
> Native GB can handle paste of markdown (markdown -> blocks). Will this do the same? 👀

Not at the moment, but I could look into it.

dingo-d (dismissed):

blocks/init/src/Blocks/components/paragraph/components/paste-handling.js
```diff
@@ -0,0 +1,136 @@
...
+            handlePaste(inputText, attributes, setAttributes);
+        });
+    }
+    // ...else handle plaintext
+    else if (copiedData) {
+        copiedData.getAsString((inputText) => handlePaste(inputText, attributes, setAttributes));
+    }
+};
+
+export { pasteInto };
```
iruzevic, line 136: is this something that is going to change to be project specific or is this going to be always this way in any project?

goranalkovic-infinum: That part should stay the same, it uses configurable options from the manifest (allowed tags) if stuff needs to be changed in a project.

iruzevic: then this should go to https://github.com/infinum/eightshift-frontend-libs/tree/develop/scripts/editor and be exported and used directly from the lib and not copied to the project

scripts/helpers/index.js
```diff
@@ -9,3 +9,4 @@ export { selector } from './selector';
 export { checkAttr } from './check-attr';
 export { elementChildrenHeight } from './element-children-height';
 export { camelize } from './camelize';
+export { pasteInto } from './paste-handler';
```
iruzevic, line 12: I think that this should go to https://github.com/infinum/eightshift-frontend-libs/tree/develop/scripts/editor because it is only editor related helper

iruzevic (changes requested):

dingo-d (approved):

gabriel-glo (approved):

iruzevic (approved):

PR #280, opened 2021-02-10
Icon updates

### LINKED ISSUE #264: Missing block icons
* Email
* Shortcode
* FeaturedCategoriesPosts

Also can you make few card variations like 5-6 so we can use them on different projects

goranalkovic-infinum:
Closes #264.

# Changes
- added new text align icons
- added new block icons
- converted existing block icons to be in string format (and use non-JSX-style properties - `fill-rule` instead of `fillRule`)
- blocks can now pull icons from the internal library
- updated blocks to use name-based icons
- added slight background to block icons

## Name-based library icons

To use library icons in blocks in manifest write:
```json
"icon": {
	"name": "example"
},
```
If the given name doesn't exist, a fallback icon is used (`blockIcons.example`)

## New icon background

image

## New icons
UI icons:
image

Block icons:
image


dingo-d (dismissed):

iruzevic (dismissed):

IvanGrginovInf:
This is awesome and you're awesome @goranalkovic-infinum . One small thing that nags me with this. It's potentially going to be confusing / difficult to debug trying to use Dashicons vs our icons. For example:

Using dashicons:
```js
  "icon": {
    "src": "editor-rtl"
  },
```

Using our icons:
```js
  "icon": {
    "name": "accordion"
  },
```

I can forsee some headbanging of the "why is my icon not loading" nature because you're using `src` and trying to load our own icons or using `name` and trying to load Dashicons.

IvanGrginovInf (dismissed):

goranalkovic-infinum:
> I can forsee some headbanging of the "why is my icon not loading" nature because you're using `src` and trying to load our own icons or using `name` and trying to load Dashicons.

That sounds like a bad time.

First I was considering implementing our custom icon to be a separate key (something like `"eightshiftIcon": "name"`), but that was kinda clunky to me.

Then I tried making it work with just `src`, but unfortunately it always overrides it to dashicon if a string is present in `src`.

That's how I landed at this implementation.

If you have any ideas how to make it less confusing and easier to debug, I'll gladly implement it!

Maybe I could add a fallback where if there are no eightshift icons for that string, a dashicon would be loaded instead?

dingo-d:
Why not prefixing the icon names? That's what font awesome always did. You had: `fa-user`, or something like that. Our iconset could be `es-card`.

IvanGrginovInf:
> Then I tried making it work with just src, and made it work, but then it would be even more confusing - do you prioritize eightshift icons or dashicons? do you prefix eightshift icons not to mix them with dashicons?

> Why not prefixing the icon names? That's what font awesome always did. You had: fa-user, or something like that. Our iconset could be es-card.

Yeah that makes sense, good idea, same key `src` we just prefix our icons.

kmaric69 (dismissed):

goranalkovic-infinum:
Thanks for the suggestions guys!

Refactored it to use `src` and added prefixes to icons. Went with `eightshift-` prefix tho, it's only used for block icons, so we don't need to save characters :D

IvanGrginovInf (approved):

iruzevic (approved):

PR #289, opened 2021-02-16
Add filter component options helper

kmaric69:
@goranalkovic-infinum and I created this helper for filtering component options.

Use case example:
You have a parent component that uses a different component for example the heading component.
The heading component has 10 different color options and 5 different size options, but you don't need all of them in your parent component.

With this helper, you can define the options you want inside the parent component and it will dynamically filter them out.

How does it work? See the Doc comment inside the helper functions. 😃

### DIFF
--- a/scripts/editor/filter-component-options.js
+++ b/scripts/editor/filter-component-options.js
@@ -1,0 +1,62 @@
+/**
+ * Use this hook to filter the component options based on the options defined in parent component
+ *
+ * Step 1:
+ * Inside the parent component define the options you want to show.
+ * Example:
+ *	"options": {
+ *		"heading": {
+ *			"colors": [
+ *				"richBlack",
+ *				"white"
+ *			],
+ *			"sizes": [
+ *				"xl",
+ *				"l"
+ *			]
+ *		}
+ * 	}
+ *
+ * Step 2:
+ * Inside the parents component options.js file where you render the component
+ * add the options you defined in parent manifest to the the componentNameOptions attribute
+ * Example:
+ * <HeadingOptions
+ * 	{...attributes}
+ * 	headingOptions={manifest.options.heading}
+ * 	setAttributes={setAttributes}
+ * />
+ *
+ * Step 3:
+ * Inside the components options.js file add the helper function.
+ *
+ * Requires WP => 5.3
+ *
+ * @param {array} attributes Attributes from the parent component.
+ * @param {array} componentName Component name from the used component.
+ * @param {array} options Options array from the component.
+ *
+ * @return object
+ *
+ */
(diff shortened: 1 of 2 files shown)


scripts/editor/filter-component-options.js
```diff
@@ -0,0 +1,62 @@
...
+ */
+export const filterComponentOptions = (attributes, componentName, options) => {
+
+	if (!attributes[`${componentName}Options`]) {
+		return;
+	}
+
+	Object.keys(options).forEach(
+		(key) => {
+			const values = options[key];
```
dingo-d, line 50: This can be placed after the conditional check where you return noting. Small performance optimization.

scripts/editor/filter-component-options.js
```diff
@@ -0,0 +1,62 @@
...
+
+			const limitedValues = attributes[`${componentName}Options`][key];
+
+			if(!limitedValues) {
+				return;
+			}
+
+			options[key] = values.filter(entry => limitedValues.includes(entry.value ?? entry));
+
+		}
```
dingo-d, line 60: You're not returning anything here, and you should if you are using arrow functions with braces.

dingo-d (changes requested):
Left some comments

scripts/editor/filter-component-options.js
kmaric69: Not sure if we need to return anything here because it's inside the forEach loop and there is nothing to return.

IvanGrginovInf (approved):

dingo-d (approved):

iruzevic (changes requested):
Awesome looks good. Please write an appropriate documentation and attach a PR here for the approval.

Add a new section here 
https://infinum.github.io/eightshift-docs/docs/basics/blocks-component-in-block

goranalkovic-infinum:
@iruzevic 
Here it is: [https://github.com/infinum/eightshift-docs/pull/77](https://github.com/infinum/eightshift-docs/pull/77) :)

iruzevic (approved):

iruzevic:
Please merge this, it is approved

kmaric69:
After deeper testing this helper I found out that it has a weird bug.

Once the helper is called and changes the options with new values, each new component that is added will inherit these values. It's like that it overrides them globaly.

I've tried to debug why this is happening but with no luck. I think it would be good if we discuss this at our next standup meeting with the whole team.

iruzevic:
This PR is no longer relevante

iruzevic:
https://github.com/infinum/eightshift-frontend-libs/pull/296 new PR

PR #292, opened 2021-02-20
Update bem menu

dingo-d:
Still needs to be tested and tweaked.

Added a JS that will make it accessible (tabbable), modified the structure to conform to the BEM structure, and work with the updated BEM menu walker.

### DIFF
--- a/blocks/init/src/Blocks/components/menu/assets/main-menu-a11y.js
+++ b/blocks/init/src/Blocks/components/menu/assets/main-menu-a11y.js
@@ -1,0 +1,109 @@
+export class AccessibleMenu {
+	constructor(options) {
+		this.menuSelector = options.menuSelector;
+		this.anchor = options.anchor;
+		this.subMenu = options.subMenu;
+
+		this.mainMenu = document.querySelector(this.menuSelector);
+		this.links = this.mainMenu.querySelectorAll(this.anchor);
+
+		this.IS_FOCUSED_CLASS = 'is-focused';
+	}
+
+	addClassName(el, cls) {
+		if (!el.className.match(`(?:^|\\s)${cls}(?!\\S)`)) {
+			el.className += ` ${cls}`;
+		}
+	}
+
+	hasClassName(el, cls) {
+		return !!el.className.match(`(?:^|\\s)${cls}(?!\\S)`);
+	}
+
+	delClassName(el, cls) {
+		el.className = el.className.replace(new RegExp(`(?:^|\\s)${cls}(?!\\S)`), '');
+	}
+
+	tabCloseMenu() {
+		const navActive = this.mainMenu.querySelectorAll(`.${this.IS_FOCUSED_CLASS}`);
+
+		for (let i = 0; i < navActive.length; i++) {
+			this.delClassName(navActive[i], this.IS_FOCUSED_CLASS);
+			navActive[i].querySelector(this.anchor).setAttribute('aria-expanded', 'false');
+		}
+	}
+
+	addAriaExpanded(element) {
+		element.setAttribute('aria-expanded', 'true');
+	}
+
+	toggleMenus(event) {
+		const currentElement = event.target;
+		const parentEl = currentElement.parentElement;
+		const insideSubMenu = currentElement.closest(this.subMenu);
+		const containsSubMenu = parentEl.querySelector(this.subMenu);
+
+		/*
+		If link contains a submenu we need to focus it,
+		if we are inside a submenu, we need to focus the parent list item.
+		Current implementation will work for level 2 submenus.
+		*/
+		if (containsSubMenu) {
+			this.addClassName(parentEl, this.IS_FOCUSED_CLASS);
+			this.addAriaExpanded(parentEl.querySelector(this.anchor));
+		}
+
+		if (insideSubMenu) {
+			const subMenuParent = insideSubMenu.parentElement;
+
+			this.addClassName(subMenuParent, this.IS_FOCUSED_CLASS);
+			this.addAriaExpanded(subMenuParent.querySelector(this.anchor));
+
+			const insideSubSubMenu = insideSubMenu.parentElement.closest(this.subMenu);
+
+			if (insideSubSubMenu) {
+				const subSubMenuParent = insideSubSubMenu.parentElement;
+
+				this.addClassName(subSubMenuParent, this.IS_FOCUSED_CLASS);
+				this.addAriaExpanded(subSubMenuParent.querySelector(this.anchor));
+			}
+		}
+
+		// Previous main menu list element (top parent).
+		const prevLi = parentEl.previousElementSibling;
+
+		if (prevLi && this.hasClassName(prevLi, this.IS_FOCUSED_CLASS)) {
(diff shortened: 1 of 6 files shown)


blocks/init/src/Blocks/components/menu/assets/main-menu-a11y.js
```diff
@@ -0,0 +1,109 @@
...
+
+			nextLi.querySelectorAll(this.anchor).forEach((anchor) => {
+				anchor.setAttribute('aria-expanded', 'false');
+			});
+		}
+	}
+
+	init() {
+		this.mainMenu.addEventListener('blur', () => this.tabCloseMenu());
+		this.mainMenu.addEventListener('mouseleave', () => this.tabCloseMenu());
```
IvanGrginovInf, line 97: "Unhover" effect should probably have a delay on it to help with accessibility, see https://www.w3.org/WAI/tutorials/menus/flyout/#fly-out-functionality

```suggestion
		this.mainMenu.addEventListener('mouseleave', setTimeout(() => this.tabCloseMenu(), 500));
```

blocks/init/src/Blocks/components/menu/assets/main-menu-a11y.js
```diff
@@ -0,0 +1,109 @@
...
+	hasClassName(el, cls) {
+		return !!el.className.match(`(?:^|\\s)${cls}(?!\\S)`);
+	}
+
+	delClassName(el, cls) {
+		el.className = el.className.replace(new RegExp(`(?:^|\\s)${cls}(?!\\S)`), '');
+	}
+
+	tabCloseMenu() {
+		const navActive = this.mainMenu.querySelectorAll(`.${this.IS_FOCUSED_CLASS}`);
```
IvanGrginovInf, line 28: Shouldn't this be `this.mainMenu.querySelector(`.${this.IS_FOCUSED_CLASS}`);`? (since there can only be 1 thing that is focused at a time).

blocks/init/src/Blocks/components/menu/assets/main-menu-a11y.js
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	addClassName(el, cls) {
+		if (!el.className.match(`(?:^|\\s)${cls}(?!\\S)`)) {
+			el.className += ` ${cls}`;
+		}
+	}
+
+	hasClassName(el, cls) {
+		return !!el.className.match(`(?:^|\\s)${cls}(?!\\S)`);
```
IvanGrginovInf, line 20:
```suggestion
		return el.classList.contains(cls);
```

blocks/init/src/Blocks/components/menu/assets/main-menu-a11y.js
```diff
@@ -0,0 +1,109 @@
...
+		this.mainMenu = document.querySelector(this.menuSelector);
+		this.links = this.mainMenu.querySelectorAll(this.anchor);
+
+		this.IS_FOCUSED_CLASS = 'is-focused';
+	}
+
+	addClassName(el, cls) {
+		if (!el.className.match(`(?:^|\\s)${cls}(?!\\S)`)) {
+			el.className += ` ${cls}`;
+		}
```
IvanGrginovInf, line 16: Why not just use `classList`? I.e. `el.classList.add(cls)`?

blocks/init/src/Blocks/components/menu/assets/main-menu-a11y.js
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	hasClassName(el, cls) {
+		return !!el.className.match(`(?:^|\\s)${cls}(?!\\S)`);
+	}
+
+	delClassName(el, cls) {
+		el.className = el.className.replace(new RegExp(`(?:^|\\s)${cls}(?!\\S)`), '');
```
IvanGrginovInf, line 24: Why not use just use `classList`? I.e. `el.classList.remove(cls)`

blocks/init/src/Blocks/components/menu/menu.php
```diff
@@ -19,20 +19,18 @@
...
 	$variation,
-	$parentClasses,
+	$linkClasses,
+	$jsModifier,
 	$modifier ? "{$variation}--{$modifier}" : ''
 );
 
 if (!empty($menu) && !$menu) {
-	echo \esc_html($menu);
+	echo $menu; // phpcs:ignore
```
IvanGrginovInf, line 35: Shouldn't we ignore specific rules here rather than blanket ignore? I.e. `// phpcs:ignore No.Escaping.Rule`

IvanGrginovInf (changes requested):
Left a few comments regarding some JS class handling, otherwise looks good!

blocks/init/src/Blocks/components/menu/assets/main-menu-a11y.js
dingo-d: 🤔  That actually makes sense 👍🏼

blocks/init/src/Blocks/components/menu/menu.php
dingo-d: Sure. I usually blanket ignore since it's easier in the projects 🤷🏼‍♂️

blocks/init/src/Blocks/components/menu/assets/main-menu-a11y.js
```diff
@@ -11,21 +11,19 @@ export class AccessibleMenu {
...
 	delClassName(el, cls) {
-		el.className = el.className.replace(new RegExp(`(?:^|\\s)${cls}(?!\\S)`), '');
+		el.classList.remove(cls);
 	}
 
 	tabCloseMenu() {
-		const navActive = this.mainMenu.querySelectorAll(`.${this.IS_FOCUSED_CLASS}`);
+		const navActive = this.mainMenu.querySelector(`.${this.IS_FOCUSED_CLASS}`);
 
 		for (let i = 0; i < navActive.length; i++) {
```
IvanGrginovInf, line 28: If you changed the above to `querySelector` now `navActive` should no longer be an array, please test it out.

dingo-d: I will test it out on the real theme in a few sec 👍🏼

IvanGrginovInf (dismissed):

dingo-d:
I'll merge this when I test this on an example project. So far the build process is broken so I'm blocked on this.

Ref. https://github.com/infinum/eightshift-libs/issues/144

goranalkovic-infinum:
Closing this, as we're not using WP menus anymore, there's a new "Site navigation" and "Site footer" block to cover basic functionality.

dingo-d:
Btw the only 'issue' with the walker in the libs was not updating the arguments (removing optional arg at the end iirc) to match the fix in the WP Core.

https://github.com/infinum/eightshift-libs/pull/373/files#diff-f7770bbdec7f1329b3dfd2de07b166174ff20669cb388f09b795e60f60ed38d7

The thing you added with site navigation and footer is messing things up for multilingual sites. 

Why didn't you start using FSE instead if you wanted to have reusable footers and headers?

PR #296, opened 2021-03-02
adding implementation to the helper

iruzevic:
Feature is the same as on this PR https://github.com/infinum/eightshift-frontend-libs/pull/289 but more streamed in the eightshift way

### DIFF
--- a/blocks/init/src/Blocks/components/button/components/button-options.js
+++ b/blocks/init/src/Blocks/components/button/components/button-options.js
@@ -4,17 +4,18 @@ import { Fragment } from '@wordpress/element';
 import { URLInput } from '@wordpress/block-editor';
 import { ColorPaletteCustom } from '@eightshift/frontend-libs/scripts/components';
 import { SelectControl, TextControl, Icon, ToggleControl } from '@wordpress/components';
-import { icons, getOptionColors } from '@eightshift/frontend-libs/scripts/editor';
+import { icons, getOptionColors, getOptions } from '@eightshift/frontend-libs/scripts/editor';
 import { checkAttr } from '@eightshift/frontend-libs/scripts/helpers';
 import manifest from './../manifest.json';
 
-const { options, title } = manifest;
+const { title } = manifest;
 
 export const ButtonOptions = (attributes) => {
 	const {
 		setAttributes,
 		componentName = manifest.componentName,
 		label = title,
+		options = options,
 		buttonShowControls = true,
 
 		buttonUse = checkAttr('buttonUse', attributes, manifest, componentName),
@@ -81,7 +82,7 @@ export const ButtonOptions = (attributes) => {
 								</Fragment>
 							}
 							value={buttonColor}
-							colors={getOptionColors(options.colors)}
+							colors={getOptionColors(getOptions(manifest, componentName, 'color', options))}
 							onChange={(value) => setAttributes({ [`${componentName}Color`]: value })}
 						/>
 					}
@@ -95,7 +96,7 @@ export const ButtonOptions = (attributes) => {
 								</Fragment>
 							}
 							value={buttonSize}
-							options={options.sizes}
+							options={getOptions(manifest, componentName, 'size', options)}
 							onChange={(value) => setAttributes({ [`${componentName}Size`]: value })}
 						/>
 					}
@@ -109,7 +110,7 @@ export const ButtonOptions = (attributes) => {
 								</Fragment>
 							}
 							value={buttonWidth}
-							options={options.widths}
+							options={getOptions(manifest, componentName, 'width', options)}
 							onChange={(value) => setAttributes({ [`${componentName}Width`]: value })}
 						/>
 					}
--- a/blocks/init/src/Blocks/components/button/components/button-toolbar.js
+++ b/blocks/init/src/Blocks/components/button/components/button-toolbar.js
@@ -3,15 +3,17 @@ import { __, sprintf } from '@wordpress/i18n';
 import { AlignmentToolbar } from '@wordpress/block-editor';
 import { Fragment } from '@wordpress/element';
 import { checkAttr } from '@eightshift/frontend-libs/scripts/helpers';
+import { getOptions } from '@eightshift/frontend-libs/scripts/editor';
 import manifest from './../manifest.json';
 
-const { options, title } = manifest;
+const { title } = manifest;
 
 export const ButtonToolbar = (attributes) => {
 	const {
 		setAttributes,
 		componentName = manifest.componentName,
 		label = title,
+		options = options,
 		buttonShowControls = true,
 
 		buttonUse = checkAttr('buttonUse', attributes, manifest, componentName),
@@ -31,7 +33,7 @@ export const ButtonToolbar = (attributes) => {
 					{showButtonAlign &&
 						<AlignmentToolbar
 							value={buttonAlign}
-							options={options.aligns}
+							options={getOptions(manifest, componentName, 'align', options)}
 							label={sprintf(__('%s text align', 'eightshift-frontend-libs'), label)}
 							onChange={(value) => setAttributes({ [`${componentName}Align`]: value })}
 						/>
--- a/blocks/init/src/Blocks/components/button/manifest.json
+++ b/blocks/init/src/Blocks/components/button/manifest.json
@@ -62,7 +62,7 @@
 		}
 	},
 	"options": {
-		"sizes": [
+		"buttonSize": [
 			{
 				"label": "Default",
 				"value": "default"
@@ -72,7 +72,7 @@
 				"value": "big"
 			}
 		],
-		"widths": [
+		"buttonWidth": [
 			{
 				"label": "Default",
 				"value": "default"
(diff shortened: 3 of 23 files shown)


kmaric69 (approved):

blocks/init/src/Blocks/components/button/components/button-options.js
```diff
@@ -4,17 +4,18 @@ import { Fragment } from '@wordpress/element';
 import { URLInput } from '@wordpress/block-editor';
 import { ColorPaletteCustom } from '@eightshift/frontend-libs/scripts/components';
 import { SelectControl, TextControl, Icon, ToggleControl } from '@wordpress/components';
-import { icons, getOptionColors } from '@eightshift/frontend-libs/scripts/editor';
+import { icons, getOptionColors, getOptions } from '@eightshift/frontend-libs/scripts/editor';
 import { checkAttr } from '@eightshift/frontend-libs/scripts/helpers';
 import manifest from './../manifest.json';
 
-const { options, title } = manifest;
+const { title } = manifest;
```
IvanGrginovInf, line 11: If you're doing this outside of the component (since it's outside of `ButtonOptions`), doesn't this automatically make the variable global?

iruzevic: no sure, we didn't have any problems on multiple block. Generally my understanding of JS is that there is no such thing as global unless you put it in a window object. Everything else is scoped by babel and webpack

blocks/init/src/Blocks/components/button/components/button-options.js
```diff
@@ -81,7 +82,7 @@ export const ButtonOptions = (attributes) => {
 								</Fragment>
 							}
 							value={buttonColor}
-							colors={getOptionColors(options.colors)}
+							colors={getOptionColors(getOptions(manifest, componentName, 'color', options))}
```
dingo-d, line 85: I think it would be more readable if all these `getOptions` calls were extracted out of the `return` code 🤷🏼‍♂️ 

This applies to the changes on lines 99 and 113 as well.

blocks/init/src/Blocks/components/button/components/button-toolbar.js
```diff
@@ -31,7 +33,7 @@ export const ButtonToolbar = (attributes) => {
 					{showButtonAlign &&
 						<AlignmentToolbar
 							value={buttonAlign}
-							options={options.aligns}
+							options={getOptions(manifest, componentName, 'align', options)}
```
dingo-d, line 36: ⬆️

blocks/init/src/Blocks/components/button/manifest.json
```diff
@@ -62,7 +62,7 @@
 		}
 	},
 	"options": {
-		"sizes": [
+		"buttonSize": [
```
dingo-d, line 65: Does this have to be prefixed with the `button`? Seems kinda redundant as we are in the button manifest 👀

blocks/init/src/Blocks/components/heading/components/heading-options.js
```diff
@@ -54,7 +55,7 @@ export const HeadingOptions = (attributes) => {
 									{__('Color', 'eightshift-frontend-libs')}
 								</Fragment>
 							}
-							colors={getOptionColors(options.colors)}
+							colors={getOptionColors(getOptions(manifest, componentName, 'color', options))}
```
dingo-d, line 58: Same comment as others about readability.

blocks/init/src/Blocks/components/heading/components/heading-options.js
```diff
@@ -69,7 +70,7 @@ export const HeadingOptions = (attributes) => {
 								</Fragment>
 							}
 							value={headingSize}
-							options={options.sizes}
+							options={getOptions(manifest, componentName, 'size', options)}
```
dingo-d, line 73: ⬆️

blocks/init/src/Blocks/components/heading/components/heading-toolbar.js
```diff
@@ -42,7 +44,7 @@ export const HeadingToolbar = (attributes) => {
 					{showHeadingAlign &&
 						<AlignmentToolbar
 							value={headingAlign}
-							options={options.aligns}
+							options={getOptions(manifest, componentName, 'align', options)}
```
dingo-d, line 47: ⬆️

blocks/init/src/Blocks/components/heading/manifest.json
```diff
@@ -39,7 +39,7 @@
 		}
 	},
 	"options": {
-		"sizes": [
+		"headingSize": [
```
dingo-d, line 42: Same as the above

blocks/init/src/Blocks/components/button/components/button-options.js
IvanGrginovInf: ehh, still it can be done couple of lines below inside the component where you do `const { options } = manifest;`, doesn't hurt

blocks/init/src/Blocks/components/button/components/button-options.js
iruzevic: sure but I would like to have all the component props at one place so you can reuse them easily

blocks/init/src/Blocks/components/button/manifest.json
iruzevic: Yes, because without that this wont work. The problem is that you can have options name buttonSize or introSize because of the components in manifest. Without this you would need to map additional key in the helper. And with this way you know exactly for that this options is used

blocks/init/src/Blocks/components/button/components/button-options.js
iruzevic: this is just wild guessing. If it starts making problems I will move it

dingo-d (changes requested):
Left some comments.

I'm a bit at a loss how and what this does exactly, but we'll have a documentation 😄 

Also, don't forget to add @kmaric69 as a co-author 😉

blocks/init/src/Blocks/components/button/components/button-options.js
iruzevic: I have asked JS guys

dingo-d: I have this in all my components, didn't see any issues so far. Then again, I never had to override some parent options 🤔  What if this was causing issues for @kmaric69 ?

blocks/init/src/Blocks/components/button/components/button-options.js
dingo-d: Not sure I understand. What would be different if the `getOptions(manifest, componentName, 'color', options)` was set as `const colorOptions` and then passed here?

blocks/init/src/Blocks/components/button/manifest.json
dingo-d: Kinda ugly, but makes sense xD

blocks/init/src/Blocks/components/button/components/button-options.js
iruzevic: if you have a big component you need to scroll up and down to see the configuration and what it does. Also all component options are scoped under the condition so this way this helper will not fire if it is not in use. 

Your way you need to put a variable on the top, scroll up and down to read what component does and you need to also put a if condition on the variable on the top so you duplicate your work.

blocks/init/src/Blocks/components/button/components/button-options.js
iruzevic: Kristijan had different issues with his helper

IvanGrginovInf (approved):

blocks/init/src/Blocks/components/button/components/button-options.js
IvanGrginovInf: Just saying, there's no need to do it outside of the component (outside of `export const ButtonOptions = (attributes) => {`) and while it might not cause issues, it's a bit confusing. But if it's not causing issues I guess it's ok, will approve.

IvanGrginovInf (approved):

blocks/init/src/Blocks/components/button/components/button-options.js
dingo-d: > Also all component options are scoped under the condition so this way this helper will not fire if it is not in use.

If you have just one place where you use it, sure, but if you have it in multiple places chances are that you'll use it. It's imported anyhow, so I don't think it will be huge a performance loss.

> Your way you need to put a variable on the top, scroll up and down to read what component does and you need to also put a if condition on the variable on the top so you duplicate your work.

Wrapper has this, tons of spacing options are defined at the top, and used below in the components. I really don't have a big problem with this.

I'll always choose duplication over the loss of readability tbh 🤷🏼‍♂️  

I'm fine to leave it in, it just looks super hard to understand what is happening here, especially since it's wrapped in `getOptionColors` helper.

dingo-d:
You didn't add `Co-authored-by: name <name@example.com>` for Kiki. Since he had a PR off of which you've worked off he should have gotten some credit.

Also, my [comment](https://github.com/infinum/eightshift-frontend-libs/pull/296#discussion_r585467053) was just overlooked (without any comment from anybody from the team), which is kinda not cool...

PR #298, opened 2021-03-02
style updates

iruzevic:
- converting px to em
- converting <Fragment> to <>
- added advance class from native core to wrapper by default


blocks/init/src/Blocks/components/accordion/accordion-style.scss
```diff
@@ -1,12 +1,12 @@
...
 	border-color: global-settings(colors, light),
-	icon-size: 30px,
+	icon-size: 1.5em,
 );
 
 .accordion {
 	$this: &;
 	position: relative;
-	border-bottom: 1px solid map-get-strict($accordion, border-color);
+	border-bottom: 0.05em solid map-get-strict($accordion, border-color);
```
goranalkovic-infinum, line 9: We should consider using `rem` instead of `em` in places like these, we wouldn't want the border thickness to vary if a div that the accordion is in has a bigger base font size.

goranalkovic-infinum (commented):
We should consider using `rem` (uses body/html font-size as base for calculation) instead of `em` (uses font-size inherited from the element that it's contained inside as base for calculation) in places like border thicknesses, absolute positionings for some items, etc.

Do we want the border thickness to vary if a div that the block is in has a bigger base font size?

IvanGrginovInf (approved):

kmaric69 (approved):

dingo-d:
Good point. I've had ems sometimes bite me in the ass 😬  Love relative units, but we need to be careful when we want to use `em` and when to use `rem`.

blocks/init/src/Blocks/components/heading/manifest.json
```diff
@@ -41,11 +41,11 @@
 	"options": {
 		"headingSize": [
 			{
-				"label": "Default (30px)",
+				"label": "Default (1.5em)",
 				"value": "default"
 			},
 			{
-				"label": "Big (50px)",
+				"label": "Big (2.5em)",
```
dingo-d, line 48: Maybe we should remove units from the labels? Pixels kinda made sense to devs (more traditional), but to clients, the ems won't make any sense.

blocks/init/src/Blocks/components/link/manifest.json
```diff
@@ -60,11 +60,11 @@
 	"options": {
 		"linkSize": [
 			{
-				"label": "Default (22px)",
+				"label": "Default (1.1em)",
 				"value": "default"
 			},
 			{
-				"label": "Small (18px)",
+				"label": "Small (0.9em)",
```
dingo-d, line 67: ⬆️

blocks/init/src/Blocks/components/lists/manifest.json
```diff
@@ -41,11 +41,11 @@
 	"options": {
 		"sizes": [
 			{
-				"label": "Default (22px)",
+				"label": "Default (1.1em)",
 				"value": "default"
 			},
 			{
-				"label": "Small (18px)",
+				"label": "Small (0.9em)",
```
dingo-d, line 48: ⬆️

blocks/init/src/Blocks/components/paragraph/manifest.json
```diff
@@ -36,11 +36,11 @@
 	"options": {
 		"paragraphSize": [
 			{
-				"label": "Default (22px)",
+				"label": "Default (1.1em)",
 				"value": "default"
 			},
 			{
-				"label": "Small (18px)",
+				"label": "Small (0.9em)",
```
dingo-d, line 43: ⬆️

dingo-d (changes requested):
Left some inline comments plus we should discuss the issue from @goranalkovic-infinum 
 comment.

blocks/init/src/Blocks/components/heading/manifest.json
goranalkovic-infinum: Agreed. Non-dev people would very likely be very confused with ems. We should either leave the pixel values or just remove the units altogether.

iruzevic: I will fix this in another PR

blocks/init/src/Blocks/components/link/manifest.json
iruzevic: I will fix this in another PR

blocks/init/src/Blocks/components/lists/manifest.json
iruzevic: I will fix this in another PR

PR #302, opened 2021-03-08
adding option for css variables

iruzevic:
This is just a little update for webpack. Nothing to do with css variables

### DIFF
--- a/webpack/helpers.js
+++ b/webpack/helpers.js
@@ -140,24 +140,24 @@ function convertJsonToSassVariables(data) {
 		if (Object.prototype.hasOwnProperty.call(data, property)) {
 			switch (typeof data[property]) {
 				case 'object':
-						for (const inner in data[property]) {
-							if (Object.prototype.hasOwnProperty.call(data[property], inner)) {
-								switch (property) {
-									case 'colors':
-										output += `--${property}-${data[property][inner].slug}: ${data[property][inner].color};\n`;
-										break;
-									case 'gradient':
-										output += `--${property}-${data[property][inner].slug}: ${data[property][inner].gradient};\n`;
-										break;
-									case 'fontSizes':
-										output += `--${property}-${data[property][inner].slug}: ${data[property][inner].slug};\n`;
-										break;
-									default:
-										output += `--${property}-${inner}: ${data[property][inner]};\n`;
-										break;
-								}
+					for (const inner in data[property]) {
+						if (Object.prototype.hasOwnProperty.call(data[property], inner)) {
+							switch (property) {
+								case 'colors':
+									output += `--${property}-${data[property][inner].slug}: ${data[property][inner].color};\n`;
+									break;
+								case 'gradient':
+									output += `--${property}-${data[property][inner].slug}: ${data[property][inner].gradient};\n`;
+									break;
+								case 'fontSizes':
+									output += `--${property}-${data[property][inner].slug}: ${data[property][inner].slug};\n`;
+									break;
+								default:
+									output += `--${property}-${inner}: ${data[property][inner]};\n`;
+									break;
 							}
 						}
+					}
 					break;
 				default:
 					output += `--${property}: ${data[property]};\n`;
--- a/webpack/helpers.js
+++ b/webpack/helpers.js
@@ -4,6 +4,7 @@
  */
 
 const path = require('path');
+const fs = require('fs');
 
 /**
  * Generate all paths required for Webpack build to work.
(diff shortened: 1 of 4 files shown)


webpack/helpers.js
```diff
@@ -125,13 +128,63 @@ function convertJsonToSassGeneral(data) {
...
+ * @param {object} data Json data to convert.
+ */
+function convertJsonToSassVariables(data) {
+	let output = '';
+
+	for (const property in data) {
+		if (Object.prototype.hasOwnProperty.call(data, property)) {
+			switch (typeof data[property]) {
+				case 'object':
+						for (const inner in data[property]) {
```
goranalkovic-infinum, line 143: Did this go 1 tab too far right?

iruzevic: yep it did. Fixed

goranalkovic-infinum (commented):
Awesome, but I feel like we should filter which things we convert to variables - do we really need stuff like component name and breakpoints defined in every block? It kinda defies the advantage of CSS variables.

iruzevic:
> Awesome, but I feel like we should filter which things we convert to variables - do we really need stuff like component name and breakpoints defined in every block? It kinda defies the advantage of CSS variables.
> 
> Also, stuff like colors and breakpoints should be defined globally, then if you need to change it inside of a component/block, you can override the variable in per-use-case basis.

This are all variables that are global. Just like before we used them like global-settings helper and now you will have them as css variables. Before you have helper for maps now I will create a helper that creates root from all this variables so thay are available.

goranalkovic-infinum:
> This are all ...

Awesome!

One more nitpick before approving, could we switch things to be snake-case for all of the variables?
(e.g. `sectionSpacing` to `section-spacing`)

webpack/helpers.js
```diff
@@ -125,13 +128,63 @@ function convertJsonToSassGeneral(data) {
...
+							}
+						}
+					}
+					break;
+				default:
+					output += `--${property}: ${data[property]};\n`;
+					break;
+			}
+		}
+	}
```
dingo-d, line 167: There isn't any prettier way to write this? Like some helper method or something. Loop then if, then switch, then another loop... Looks messy :S

kmaric69 (dismissed):

iruzevic:
> > This are all ...
> 
> Awesome!
> 
> One more nitpick before approving, could we switch things to be snake-case for all of the variables?
> (e.g. `sectionSpacing` to `section-spacing`)

at this point no because this will break back compatibility. we will change this in the later major release. In general i don't do any manipulations here. How it is defined in the manifest it is outputted here. and the variables are camelcased because of the usage in the JS.

We can convert all to camelcase for variables like:
```css
--breakpointsMobile: 767;
--sectionSpacingMin: -300;
```

IvanGrginovInf (dismissed):

dingo-d (dismissed):
Left one comment.

For rationale why I consider this 'ugly' check this post: http://blog.cleancoder.com/uncle-bob/2021/03/06/ifElseSwitch.html

iobrado (dismissed):

webpack/helpers.js
goranalkovic-infinum: This could be way more elegant with a simple if else:

```js
if (typeof data[property] !== 'object') {
    output += `--${property}: ${data[property]};\n`;
    break;
}

for (const inner ...)
...
```

Also, calling `Object.prototype.hasOwnProperty` could be avoided with just going through the `Object.entries` in a `for ... of ...` loop, because then you're sure that the object has that property.

Removes tons of indentation 😄

volfkarlo:
I would say that constants like.

`--customBlocksName: eightshift-block;
--maxCols: 12;
--breakpoints-mobile: 767;
--breakpoints-tablet: 991;
--breakpoints-desktop: 1199;`

don't have to be used as CSS variables. Because it seems to me like we are using them for sake of their usage. 
I would say that we should define variables in places where you have to reuse or redefine some variables. For example

```
.transition-delay--200 {
--transition-delay: 200ms;
}
.transition-delay--300 {
--transition-delay: 300ms;
}
.transition-delay--400 {
--transition-delay: 400ms;
}

.fade-in {
  @foreach $i from 1 through 100 {
       &:nth-child(#{$i}) {
          transition-delay: #{$i} * var(--transition-delay) + 'ms';
       }
    }
}

``` 

In this case we won't have to generate `delayModifiers * numberOfChildren` classes but `numberOfChildren + delayModifiers`. And I believe that is the purpose of css variables.

@iruzevic what do you say about that?

IvanGrginovInf (approved):
Approved but please fix these linting errors:

`Expected to return a value at the end of method 'requestToExternal'`

dingo-d (approved):

PR #304, opened 2021-03-17
Dart Sass implementation

gabriel-glo:
Dart Sass is the preferred and primary method for implementing Sass. LibSass and Node Sass are deprecated.
Node Sass is using `node-gyp` package (that requires x-code tools) and is known to cause issues periodically.

`sass-loader` package is already set to work with either `sass` or `node-sass`.

---

I had to add fixes to the modifiers mixins because there were some warnings about interpolation of color names.

```bash
WARNING: You probably don't mean to use the color value white in interpolation here.
It may end up represented as white, which will likely produce invalid CSS.
Always quote color names when using them as strings or map keys (for example, "white").
If you really want to use the color value here, use '"" + $modifier'.

   ╷
45 │             &--#{$modifier} {
   │                  ^^^^^^^^^
   ╵
    styles/scss/mixins/_modifiers-deep.scss 45:9                           modifiers-deep()
    blocks/init/src/Blocks/components/paragraph/paragraph-style.scss 40:3  @import
    blocks/init/src/Blocks/assets/styles/application-blocks.scss 38:9      root stylesheet
```

Solution is presented in the warning text. Other solution would be to change color names (white, black, gray .etc)

https://stackoverflow.com/questions/57455245/how-to-i-solve-this-sass-interpolation-issue/62371464#62371464


iruzevic (approved):
if everything works awesome

gabriel-glo:
I ran the tests (both unit and integration) locally and everything passed. I'm not sure what is failing on GH Actions. Are they the same tests?

Also I did some of my own tests on the button and some other components. Changed a bunch of properties just to see if it compiles correctly and there were no issues. 
It would be great if someone could temporarily test this branch on their own project.

dingo-d:
So, this will work if you have `node-sass`, and `dart-sass`?

kmaric69 (approved):

gabriel-glo:
Yes, you are controlling which implementation will be used in `sass-loader` option: `implementation: require("sass"),`.

IvanGrginovInf:
@gabriel-glo and this works without any external dependencies you need to have on your machine? Also are there any significant changes in build times?

gabriel-glo:
@IvanGrginovInf 

I didn't install anything other than `sass` package. In the documentation there is no mention of any dependency. It is using JS API to compile scss files.
https://github.com/sass/dart-sass/blob/master/README.md#javascript-api

As for the build times, I didn't do the profiling, but subjectively it seems faster on the Storybook build.

IvanGrginovInf:
> As for the build times, I didn't do the profiling, but subjectively it seems faster on the Storybook build. What do you suggest for measuring this?

In that case it's fine, we're not strict about the build times and if it doesn't feel significantly slower I'm 100% okay with it, gj!

IvanGrginovInf (approved):

iruzevic:
Yea I have read that they are dropping node-sass support and that we should transfer to this. so good work

dingo-d (approved):

goranalkovic-infinum (approved):
🎉

PR #306, opened 2021-03-19
Feature/props helper

iruzevic:
- new super mega uber duper helper for passing props. No more props spreading and need to map keys.
- load time optimisations for components registration.
- some typos
- creating global window object with all the data for faster cached usage. 
- building dependency tree for components and blocks.
- I will be sending the php equivalent tommorow

### DIFF
--- a/scripts/editor/index.js
+++ b/scripts/editor/index.js
@@ -16,3 +16,4 @@ export {
 	outputCssVariables,
 	getUnique
 } from './output-css-variables';
+export { props } from './props';
--- a/scripts/editor/props.js
+++ b/scripts/editor/props.js
@@ -1,15 +1,15 @@
 /**
  * Output only attributes that are used in the component and remove everything else.
  *
- * @param {object} globalManifest Global manifest object to get the namespace value.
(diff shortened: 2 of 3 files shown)


dingo-d (commented):
There is a test failin and some linter issues, can those be addressed?

Also the name props could be confusing, since react is always referencing props and all

iruzevic:
> There is a test failin and some linter issues, can those be addressed?
> 
> Also the name props could be confusing, since react is always referencing props and all

I have linted all the code and there are no errors showing.

Regarding the tests these are part of the create-project-script and I didn't touch that. @IvanGrginovInf can you take a look.

iruzevic:
> There is a test failin and some linter issues, can those be addressed?
> 
> Also the name props could be confusing, since react is always referencing props and all

Regarding the props name I have named it like that because it does just that manipulations on props

dingo-d (dismissed):

gabriel-glo (dismissed):

gabriel-glo (approved):

goranalkovic-infinum (approved):

PR #308, opened 2021-03-25
Improved SSR component

goranalkovic-infinum:
Adds the improved `ServerSideRender` component [by David Gwyer](https://github.com/dgwyer/server-side-render-x).


iruzevic (dismissed):

scripts/components/server-side-render/server-side-render.js
```diff
@@ -0,0 +1,170 @@
...
+	LoadingResponsePlaceholder: ({ className }) => {
+		return (
+			<Placeholder className={className}>
+				<Spinner />
+			</Placeholder>
+		);
+	},
+};
+
+export default ServerSideRender;
```
dingo-d, line 170: It says no newline at the end. Can you configure your vscode to automatically add newlines at the end of the file on saving?

Why: https://unix.stackexchange.com/a/18789
How: https://stackoverflow.com/a/44704969/629127

dingo-d (changes requested):
One minor change needed

scripts/components/server-side-render/server-side-render.js
goranalkovic-infinum: Didn't know that was a setting, thanks!

iruzevic (approved):

dingo-d (approved):

PR #311, opened 2021-04-08
updating helper for css variables

iruzevic:
adding options to convert css variable output to color, select-variable, boolean-variable and default. also option to add any css variable no mater the manifest

### DIFF
--- a/scripts/editor/output-css-variables.js
+++ b/scripts/editor/output-css-variables.js
@@ -99,8 +99,16 @@ export const outputCssVariables = (attributes, manifest, unique) => {
 
 		let innerValue = value;
 
-		if (_.has(manifest['attributes'][key], 'color')) {
+		if (manifest['attributes'][key]['variable'] === 'color') {
 			innerValue = `var(--global-colors-${innerValue})`;
+		}
+
+		if (manifest['attributes'][key]['variable'] === 'select-variable' && _.has(manifest['options'], key)) {
+			innerValue = manifest['options'][key].filter((item) => item.value === attributes[key])[0].variable;
+		}
+
+		if (manifest['attributes'][key]['variable'] === 'boolean-variable' && _.has(manifest['options'], key) && _.has(manifest['options'][key], Number(attributes[key]))) {
+			innerValue = manifest['options'][key][Number(attributes[key])];
 		}
 
 		const innerKey = _.kebabCase(key);
@@ -111,6 +119,7 @@ export const outputCssVariables = (attributes, manifest, unique) => {
 	return <style dangerouslySetInnerHTML={{__html: `
 		.${name}[data-id='${unique}'] {
 			${output}
+			${_.has(manifest, 'variables') ? manifest['variables'].join(";\n") : ''}
 		}
 	`}}></style>;
 }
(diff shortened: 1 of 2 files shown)


scripts/editor/output-css-variables.js
```diff
@@ -99,18 +99,34 @@ export const outputCssVariables = (attributes, manifest, unique) => {
...
 		}
 
+		// Output select variable from the options array but dont use value key. It will use variable key.
+		if (manifest['attributes'][key]['variable'] === 'select-variable' && _.has(manifest['options'], key)) {
+			const selectVariable = manifest['options'][key].filter((item) => item.value === attributes[key])[0].variable;
+			innerValue = typeof selectVariable === 'undefined' ? attributes[key] : selectVariable;
+		}
+
+		// Output boolean variable from the options array key. First key is false value, second is true value.
+		if (manifest['attributes'][key]['variable'] === 'boolean-variable' && _.has(manifest['options'], key) && manifest['options'][key].length === 2) {
```
dingo-d, line 114: I'd put the has check first, then the rest

dingo-d (changes requested):
One small comment

goranalkovic-infinum (approved):

scripts/editor/output-css-variables.js
```diff
@@ -99,18 +99,34 @@ export const outputCssVariables = (attributes, manifest, unique) => {
```
volfkarlo, line 99: I couldn't figure out if this part of the code has an option to reset the value to none. It should be something like

```suggestion
if (value === undefined) {
   continue;
}
```

This is quite useful for resetting the values when responsive layout should inherit something from its parent. I could take, as an example wrapper with full width. Let's say that we manually set 
```
{
  large: true,
  desktop: true,
  tablet: false,
  mobile: false,
}
```

but to generate fewer CSS variables this could be set from admin as
```
{
  large: true,
  tablet: false,
}
```
if we set the value to undefined we could achieve this 😄

volfkarlo (changes requested):
Just a minor suggestion, basically something that I've been working on locally but couldn't fully figure out how to do it since the only place I needed exceptions were inside of a wrapper. Good job! 👍 🎉

iruzevic:
Implemented all corrections.

scripts/editor/output-css-variables.js
```diff
@@ -97,20 +97,40 @@ export const outputCssVariables = (attributes, manifest, unique) => {
...
-		if (_.has(manifest['attributes'][key], 'color')) {
+		// Output color variable from the global variables.
+		if (manifest['attributes'][key]['variable'] === 'color') {
 			innerValue = `var(--global-colors-${innerValue})`;
 		}
 
+		// Output select variable from the options array but dont use value key. It will use variable key.
+		if (_.has(manifest['options'], key) && manifest['attributes'][key]['variable'] === 'select-variable') {
+			const selectVariable = manifest['options'][key].filter((item) => item.value === attributes[key])[0].variable;
+			innerValue = typeof selectVariable === 'undefined' ? attributes[key] : selectVariable;
```
goranalkovic-infinum, line 114: Could this be written as `innerValue = selectVariable ?? attributes[key]` ?

iruzevic: no sure, but I would leave it like this so it is the same as with PHP for easier debugging.

goranalkovic-infinum (approved):

dingo-d (approved):

volfkarlo (approved):

PR #317, opened 2021-04-15
Allow overriding bp branch in plugin setup

IvanGrginovInf:


iruzevic (dismissed):

dingo-d (commented):
One test is failing on node v15, you could set this as an allowed failure in the workflows?

IvanGrginovInf:
> One test is failing on node v15, you could set this as an allowed failure in the workflows?

I just fixed that yesterday by setting the note version as `15.x` rather than `15` in the matrix. But now it's happening again, not sure why. It fails with `npm ERR! Cannot destructure property 'name' of 'node' as it is null.` on `npm install` 🙈

IvanGrginovInf:
@dingo-d guess I fixed it 😅

dingo-d (approved):

iruzevic (approved):

PR #318, opened 2021-04-15
Link toolbar button

### LINKED ISSUE #181: [FEATURE] Improve the button component
Currently, the 'button' component is always a link (`a`). We should make a distinction whether we want this to be a link (URL provided) or just a button component that can be used in forms etc. (HTML `<button>` tag).

Besides that, if used as a link, new options should be added:

- `target` attribute with preset values in a dropdown
  - `_blank`, `_self`, `_parent`, `_top`
- `noreferrer` and `noopener` toggle (like in the core)
  - this should be researched a bit as it could have a SEO impact (https://twitter.com/jdevalk/status/1322300284748681216?s=20)

goranalkovic-infinum:
Closes #181.

This PR replaces the ugly old `URLInput` in the option with the fresh new `LinkInput`. As `LinkInput` itself wasn't meant to be used in an option panel, the URL picker now lies in the block toolbar (just like in stock Gutenberg).

To easily accomplish this (as there are a few elements to handle), a custom component - `LinkToolbarButton` - was created.

To use it in your blocks:
1. Import the component from FL
2. Import `useRef` and create a reference before the return function of your component
3. Add the `LinkToolbarButton` into the toolbar

Use it like this:
```jsx
<LinkToolbarButton
	componentName={componentName}
	url={buttonUrl}
	opensInNewTab={buttonOpensInNewTab}
	setAttributes={setAttributes}
	anchorRef={ref}
	title={label}
/>
```

In the editor it looks like this:
image
image

### DIFF
--- a/blocks/init/src/Blocks/components/button/components/button-options.js
+++ b/blocks/init/src/Blocks/components/button/components/button-options.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import { __, sprintf } from '@wordpress/i18n';
-import { URLInput } from '@wordpress/block-editor';
 import { ColorPaletteCustom } from '@eightshift/frontend-libs/scripts/components';
 import { SelectControl, TextControl, Icon, ToggleControl } from '@wordpress/components';
 import { icons, getOptionColors, getOptions } from '@eightshift/frontend-libs/scripts/editor';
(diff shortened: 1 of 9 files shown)


dingo-d (approved):
Awesome work 🎉

iobrado (approved):

iruzevic (approved):

PR #324, opened 2021-04-26
Icon new picker (and more CustomSelect goodies)

goranalkovic-infinum:
This PR brings 2 new things:
1. A way to render custom dropdown item and (single) value display components in `CustomSelect`
2. Enabled by 1., a new icon picker for the `icon` component (demo below)

https://user-images.githubusercontent.com/77000136/116115880-d30e7500-a6ba-11eb-8157-ca7c4ed1bf17.mov

More details in the Storybook documentation.

### DIFF
--- a/blocks/init/src/Blocks/components/icon/components/icon-options.js
+++ b/blocks/init/src/Blocks/components/icon/components/icon-options.js
@@ -9,16 +9,14 @@ import manifest from './../manifest.json';
 
 const { icons } = manifest;
 
-const IconPickerOption = props => {
-	return (
-		<components.Option {...props}>
-			<div className={'icon-option-row'}>
-				<i dangerouslySetInnerHTML={{ __html: icons[props.value] }}></i>
-				<span>{props.label}</span>
-			</div>
-		</components.Option>
-	);
-};
+const IconPickerOption = props => (
+	<components.Option {...props}>
+		<div className={'icon-option-row'}>
+			<i dangerouslySetInnerHTML={{ __html: icons[props.value] }}></i>
+			<span>{props.label}</span>
+		</div>
+	</components.Option>
+);
 
 const IconPickerValueDisplay = ({ children, ...props }) => (
 	<components.SingleValue {...props}>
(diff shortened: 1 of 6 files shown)


blocks/init/src/Blocks/components/icon/components/icon-options.js
```diff
@@ -1,9 +1,32 @@
...
+import { components } from 'react-select';
+
 import manifest from './../manifest.json';
 
+const { icons } = manifest;
+
+const IconPickerOption = props => (
+	<components.Option {...props}>
+		<div className={'icon-option-row'}>
+			<i dangerouslySetInnerHTML={{ __html: icons[props.value] }}></i>
```
dingo-d, line 15: Does this has to be `dangerouslySetInnerHTML`?

blocks/init/src/Blocks/components/icon/icon-editor.scss
```diff
@@ -0,0 +1,17 @@
...
+	grid-gap: 1rem;
+
+	i {
+		margin: 0;
+		padding: 0;
+
+		svg {
+			width: 1.2rem;
+			height: 1.2rem;
+			transform: translateY(2px);
```
dingo-d, line 14: px rem mixup 🙂 

We need to choose one and stick with it.

iruzevic (approved):

dingo-d (changes requested):
Left one minor question, and a comment. Great work!

blocks/init/src/Blocks/components/icon/components/icon-options.js
goranalkovic-infinum: If there's any better way to get a (non-JSX) SVG in string form to render inside a React component, then there is a better way (and I'll gladly change it), otherwise we need to use this 😄

blocks/init/src/Blocks/components/icon/icon-editor.scss
goranalkovic-infinum: While we did agree on using (r)ems everywhere, some things should stay in pixels, usually things like border thicknesses, shadow radii and minor transforms like this one, where you wouldn't want the base font size to affect the layout.

dingo-d (approved):

IvanGrginovInf (approved):

PR #328, opened 2021-04-29
updating new helper

iruzevic:
- updating manifest layout for custom variable key
- new helper to map responsive attributes in single object
- removing variableEditor from individual attribute. If you want to override attributes you can manual at the bottom

### DIFF
--- a/blocks/init/src/Blocks/custom/image/manifest.json
+++ b/blocks/init/src/Blocks/custom/image/manifest.json
@@ -25,43 +25,41 @@
 	},
 	"options": {
 		"imageAlign": {
-			"variable": {
-				"top left": {
-					"horizontal": "flex-start",
-					"vertical": "flex-start"
-				},
-				"top center": {
-					"horizontal": "center",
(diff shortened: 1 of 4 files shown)


goranalkovic-infinum (approved):
Awesome!

volfkarlo (commented):
>removing variableEditor from individual attribute. If you want to override attributes you can manual at the bottom

I'm not fully sure I understand it. 
Does it mean that the editor variables will also be generated for the frontend but not used or?

gabriel-glo (approved):

iruzevic:
> > removing variableEditor from individual attribute. If you want to override attributes you can manual at the bottom
> 
> I'm not sure I fully understand it.
> Does it mean that the editor variables will also be generated for the frontend but not used or?

No.the problem is adding a new set of keys for each type of variable that is used in the editor only brings so much complexity to the mix. So I removed it all together.

If you want to add some custom css variable for editor only you can use `variablesEditor` key at the bottom of the manifest

PR #329, opened 2021-04-30
Feature/check attr responsive

iruzevic:
- if you have custom key variable you must provide a custom implementation of that attribute. but if you have a value that needs attribute original value you can use %value% magic constant.

### DIFF
--- a/scripts/editor/output-css-variables.js
+++ b/scripts/editor/output-css-variables.js
@@ -177,8 +177,8 @@ export const outputCssVariables = (attributes, manifest, unique) => {
 		let value = customValue;
 
 		// If value contains magic variable swap that variable with original attribute value.
-		if (customValue === '%value%') {
-			value = originalAttribute;
+		if (customValue.includes('%value%')) {
+			value = customValue.replace('%value%', originalAttribute);
 		}
 
 		output += `--${_.kebabCase(attributeKey)}-${_.kebabCase(customKey)}: ${value};\n`;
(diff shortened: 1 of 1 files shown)


scripts/editor/output-css-variables.js
```diff
@@ -166,14 +166,22 @@ export const outputCssVariables = (attributes, manifest, unique) => {
...
-export const outputCssVariablesCustom = (objectList, attributeKey) => {
+ export const outputCssVariablesCustom = (objectList, attributeKey, originalAttribute) => {
 	let output = '';
 
 	for (const [customKey, customValue] of Object.entries(objectList)) {
-		output += `--${_.kebabCase(attributeKey)}-${_.kebabCase(customKey)}: ${customValue};\n`;
+		let value = customValue;
+
+		// If value contains magic variable swap that variable with original attribute value.
+		if (customValue === '%value%') {
```
goranalkovic-infinum, line 180:
```suggestion
		if (customValue.includes('%value%')) {
```

scripts/editor/output-css-variables.js
```diff
@@ -166,14 +166,22 @@ export const outputCssVariables = (attributes, manifest, unique) => {
...
+			value = originalAttribute;
```
goranalkovic-infinum, line 181:
```suggestion
			value = customValue.replace('%value%', originalAttribute);
```

goranalkovic-infinum (dismissed):
Left a suggestion to just replace the `%value%` inside the string instead of the whole string, in case you wanna add a prefix or suffix.

Awesome thing!

goranalkovic-infinum (approved):

PR #341, opened 2021-05-19
Optional undefinedAllowed parameter on checkAttr

volfkarlo:
Changelog:
- check-attr.js
  - added optional undefinedAllowed parameter on checkAttr function
  - removed else hance early exit

In admin options, while creating responsive attributes it appeared that we can't "force" attributes to unset the value so it can inherit from other breakpoints. That's why I added this optional parameter which can be used if it is ok for you to set values to `undefined`.

This change is only important on the admin side while setting the attributes. I believe that there is no need to make these changes on PHP part(eightshift-libs). However, if you feel like we should make it exactly the same, feel free to write to me or comment down here on PR.

Screen Shots and Screen Recordings of the use case scenario

JS additional parameter
third state of toggle

https://user-images.githubusercontent.com/46056662/118799704-53319000-b89f-11eb-9cdd-adda731fa09b.mov

detection of undefined for button

### DIFF
--- a/scripts/helpers/check-attr.js
+++ b/scripts/helpers/check-attr.js
@@ -7,41 +7,46 @@ import _ from 'lodash';
  * @param {array}  attributes Array of attributes.
  * @param {object}  manifest Array of default attributes from manifest.json.
  * @param {string} componentName The real component name.
+ * @param {boolean} undefinedAllowed Allowed detection of undefined values.
  *
  * @return mixed
  */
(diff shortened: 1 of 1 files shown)


iruzevic (approved):
Great work, I think there is no need to implement this in PHP

goranalkovic-infinum (approved):

volfkarlo:
I'll merge it. If you have any questions feel free to contact me 🙂

gabriel-glo (approved):

PR #349, opened 2021-05-26
Added editor styles for hovering on nested blocks

igrginov:
Added a visual indicator which block is hovered which helps when you need to select inner blocks.

### DIFF
--- a/blocks/init/src/Blocks/assets/styles/editor/_editor.scss
+++ b/blocks/init/src/Blocks/assets/styles/editor/_editor.scss
@@ -14,3 +14,32 @@ body.sb-show-main,
 body .interface-interface-skeleton__content {
 	font-size: var(--base-font-size);
 }
+
+/* stylelint-disable-next-line selector-max-specificity */
+.wp-block:not(.block-list-appender):not(.block-editor-default-block-appender):not(.editor-post-title):hover::before,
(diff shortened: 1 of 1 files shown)


blocks/init/src/Blocks/assets/styles/editor/_editor.scss
```diff
@@ -14,3 +14,32 @@ body.sb-show-main,
 body .interface-interface-skeleton__content {
 	font-size: var(--base-font-size);
 }
+
+/* stylelint-disable-next-line selector-max-specificity */
+.wp-block:not(.block-list-appender):not(.block-editor-default-block-appender):not(.editor-post-title):hover::before,
```
iruzevic, line 19: this should go to `override-editor.scss` file so ti is not coped in the project unnecessary 

and also everything that we do with the overriding  the editor styles must be documented in details:
* what it is used for
* why are we doing it
* what does it apply to

Because in 2-3 core updates you will have no idea why we have put this in and it will be imposible to debug and maintain.

iruzevic (changes requested):

dingo-d:
@IvanGrginovInf will this be added in the next release?

IvanGrginovInf:
Yes it can, it's just a minor improvement, not a blocker by any means

PR #358, opened 2021-05-28
Add variablesEditor to Manifest

volfkarlo:
Changelog:
- added CSS variables(variablesEditor) for only editor use
- renamed general editor only variables to variablesCustomEditor

Usage:
When there is a specific behavior that you want to override in the editor part, but keep everything in place for frontend part.
An example of behavior is under screenshots and screenrecording.

How to:
It behaves exactly the same as `variables` key from manifest. It will always override the `variables` with the same "key" because it's outputting afterward.
Screen Shot 2021-05-28 at 4 20 47 PM

Screenshots and screenrecording:
https://user-images.githubusercontent.com/46056662/119996776-58918780-bfcf-11eb-9ea4-d635e349bfb1.mov
Screen Shot 2021-05-28 at 4 13 50 PM

### DIFF
--- a/scripts/editor/output-css-variables.js
+++ b/scripts/editor/output-css-variables.js
@@ -109,6 +109,7 @@ export const outputCssVariables = (attributes, manifest, unique, globalManifest)
 	// Define variables from manifest.
 	const {
 		variables,
+		variablesEditor,
 	} = manifest;
 
 	// Get the initial data array.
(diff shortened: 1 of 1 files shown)


goranalkovic-infinum (approved):
🎉

scripts/editor/output-css-variables.js
```diff
@@ -117,49 +118,61 @@ export const outputCssVariables = (attributes, manifest, unique, globalManifest)
...
 
-			// Iterate each data array to find the correct breakpoint.
-			data.some((item, index) => {
+				// Check if we are using mobile or desktop first. Mobile first is the default.
+				const type = inverse ? 'max' : 'min';
 
-				// Check if breakpoint and type match.
-				if (item.name === breakpoint && item.type === type) {
+				// Iterate each data array to find the correct breakpoint.
+				data.some((item, index) => {
```
goranalkovic-infinum, line 154: Shouldn't this be `map` or `forEach`?

goranalkovic-infinum (changes requested):
Left one question, sorry for the early approval, didn't realize it until I opened the file by itself without changes :)

goranalkovic-infinum (approved):

scripts/editor/output-css-variables.js
volfkarlo: if I'm not wrong. This should be an optimization because it will exit as soon as the variable is inserted. 
If map/forEach was/were used it would continue through the whole array. 😄 

This might be removed/refactored when #359 is done.

goranalkovic-infinum: That's exactly right!

dingo-d (approved):

IvanGrginovInf (approved):

iruzevic (approved):

PR #360, opened 2021-06-01
Fix props logic

IvanGrginovInf:
Fixed a bug where props would throw a warning if you had a block (let's say `example`) but you don't have an `example` component in your project.

### DIFF
--- a/scripts/editor/props.js
+++ b/scripts/editor/props.js
@@ -25,17 +25,17 @@ export const props = (attributes, realName, newName = '', isBlock = false, names
 	// Get global window data.
 	const globalData = window['eightshift'][namespace].dependency;
 
-	// If component use components dependency tree.
-	// Add the current component name to the dependency array.
-	let dependency = [
-		...globalData.components[realName],
-		newNameInternal
-	];
+	let dependency = [];
 
-	// If block use blocks dependency tree.
+	// If block use blocks dependency tree, if component use components dependency tree.
 	if (isBlock) {
 		dependency = globalData.blocks[realName];
+	} else {
+		dependency = globalData.components[realName];
 	}
+
+	// Add the current component name to the dependency array.
+	dependency.push(newNameInternal);
 
 	// If you have multiple components just use one.
 	dependency = _.uniq(dependency);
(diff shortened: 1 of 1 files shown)


scripts/editor/props.js
```diff
@@ -25,18 +25,18 @@ export const props = (attributes, realName, newName = '', isBlock = false, names
...
-	// If component use components dependency tree.
-	// Add the current component name to the dependency array.
-	let dependency = [
-		...globalData.components[realName],
-		newNameInternal
-	];
+	let dependency = [];
 
-	// If block use blocks dependency tree.
+	// If block use blocks dependency tree, if component use components dependency tree.
```
goranalkovic-infinum, line 30:
```suggestion
	// If dealing with a block, use the block's dependency tree, if dealing with a component use the component's dependency tree.
```

goranalkovic-infinum (approved):
Left a minor comment

volfkarlo (approved):

iruzevic (approved):

scripts/editor/props.js
IvanGrginovInf: I mean, the original version is more concise and relays the same info 😄

goranalkovic-infinum (dismissed):

scripts/editor/props.js
```diff
@@ -25,18 +25,18 @@ export const props = (attributes, realName, newName = '', isBlock = false, names
...
-	// If component use components dependency tree.
-	// Add the current component name to the dependency array.
-	let dependency = [
-		...globalData.components[realName],
-		newNameInternal
-	];
+	let dependency = [];
 
-	// If block use blocks dependency tree.
+	// If isBlock use block's dependency tree, if component use component's dependency tree.
```
dingo-d, line 30:
```suggestion
	// If it's a block, use the block's dependency tree. If it's a component, use the component's dependency tree.
```

dingo-d (changes requested):
Left one minor comment change

volfkarlo (dismissed):

IvanGrginovInf:
@dingo-d, @goranalkovic-infinum you got me, same comment 2 times #iveBeenBested. Replaced the comment.

dingo-d (approved):

goranalkovic-infinum (approved):

volfkarlo (approved):

PR #362, opened 2021-06-07
Responsive variables

volfkarlo:
What this functionality brings is no need to rewrite the same values for each breakpoint in attributes, but to collect them as one in `variables` and reference all the breakpoints in `responsiveAttributes`. This brings us fewer points of a potential mistake(no need to enter the same code for each breakpoint), shortens the coding time(copying and correcting breakpoint name), shortens the code(images ⬇️ , wrapper 957 -> 491 LOC), and makes it altogether more simple to develop.

Specificity from the least specified to the most specified:
variables found in responsiveAttributes
variables
variablesEditor found in responsiveAttributes
variablesEditor

You can specify for example: `wrapperWidth` in `responsiveAttributes` and in `variables` but if you need something special or you need to override the default values(for something that is valid for all the other breakpoints), you can make a specific request as `wrapperWidthDesktop` and write separate definitions.
Values from `wrapperWidth` will still be defined but you can override them(images ⬇️ ).

Note:
Functionality should be ok, and backward compatibility should be completely ok.
The only concern is docs. If it is not clear please suggest for better or at least say what is not as clear as it should be.

Results screenshots:
after
before
for a specific breakpoint
after responsive attr
before responsive attr

### DIFF
--- a/scripts/editor/output-css-variables.js
+++ b/scripts/editor/output-css-variables.js
@@ -162,6 +162,8 @@ export const outputCssVariables = (attributes, manifest, unique, globalManifest)
 						// Exit.
 						return true;
 					}
+
+					return false;
 				});
 			});
 		}
@@ -259,7 +261,7 @@ const setBreakpointResponsiveVariables = (attributeVariables, breakpointName, br
 		return {
 			...attributeVariablesObject,
 			breakpoint: breakpointIndex === defaultBreakpointIndex ? 'default' : breakpointName,
-		};;
+		};
 	})
 }
 
--- a/scripts/editor/output-css-variables.js
+++ b/scripts/editor/output-css-variables.js
@@ -73,6 +73,83 @@ export const globalInner = (itemValues, itemKey) => {
 }
 
 /**
+ * Sets up a breakpoint value to responsive attribute objects from responsiveAttribute object.
+ *
+ * @param {array}  attributeVariables  - Array of attribute variables object.
+ * @param {string} breakpointName    	 - Breakpoint name from responsiveAttribute's breakpoint in block's/component's manifest.
+ * @param {number} breakpointIndex   	 - Index of responsiveAttribute's breakpoint in manifest.
+ * @param {number} numberOfBreakpoints - Number of responsiveAttribute breakpoints in block's/component's manifest.
+ *
+ * @return {array}
+ */
+const setBreakpointResponsiveVariables = (attributeVariables, breakpointName, breakpointIndex, numberOfBreakpoints) => {
+	return attributeVariables.map((attributeVariablesObject) => {
+
+		// Calculate default breakpoint index based on order of the breakpoint, inverse property and number of properties in responsiveAttributeObject.
+		const defaultBreakpointIndex = attributeVariablesObject.inverse ? 0 : (numberOfBreakpoints - 1);
+
+		// Expanding an object with an additional breakpoint property.
+		return {
+			...attributeVariablesObject,
+			breakpoint: breakpointIndex === defaultBreakpointIndex ? 'default' : breakpointName,
+		};
+	})
+}
+
+/**
+ * Iterating through variables matching the keys from responsiveAttributes and translating it to responsive attributes names.
+ *
+ * @param {object} responsiveAttributes - Responsive attributes that are read from component's/block's manifest.
+ * @param {object} variables            - Object containing objects with component's/block's attribute variables that are read from manifest.
+ *
+ * @return {object} Object prepared for setting all the variables to its breakpoints.
+ */
+const setupResponsiveVariables = (responsiveAttributes, variables) => {
+	// Iterate through responsive attributes.
+	return Object.entries(responsiveAttributes).reduce((responsiveAttributesVariables, [responsiveAttributeName, responsiveAttributeObject]) => {
+
+		// If responsive attribute doesn't exist in variables object, skip it
+		if (!responsiveAttributeName || _.isEmpty(variables[responsiveAttributeName])) {
+			return responsiveAttributesVariables;
+		}
+
+		// Used for determination of default breakpoint.
+		const numberOfBreakpoints = Object.entries(responsiveAttributeObject).length;
+
+		// Iterate each responsive attribute object as breakpoint name is the key of the object, and value represents the name of the responsive variable.
+		const responsiveAttributeVariables = Object.entries(responsiveAttributeObject).reduce((responsiveAttribute, [breakpointName, breakpointVariableName], breakpointIndex) => {
+			let breakpointVariables = {};
+
+			// Array represents direct value(default or value).
+			if (Array.isArray(variables[responsiveAttributeName])) {
+				breakpointVariables = setBreakpointResponsiveVariables(variables[responsiveAttributeName], breakpointName, breakpointIndex, numberOfBreakpoints);
+
+			// Object treatment goes depending on a value inserted(multiple choice, boolean or similar).
+			} else {
+
+				// Iterate options/multiple choices/boolean...
+				breakpointVariables = Object.entries(variables[responsiveAttributeName])
+				.reduce((acc, [attributeValue, attributeObject]) => {
+					return {
+						...acc,
+						[attributeValue]: setBreakpointResponsiveVariables(attributeObject, breakpointName, breakpointIndex, numberOfBreakpoints),
+					}
+				}, {});
+			}
+
+			// Collect all the values from one responsive attribute to one object.
+			return {
+				...responsiveAttribute,
+				[breakpointVariableName]: breakpointVariables,
+			};
+		}, {});
+
+		// Merge multiple responsive attributes to one object.
+		return {...responsiveAttributesVariables, ...responsiveAttributeVariables};
+	}, {});
+}
+
+/**
  * Setting defined variables to each breakpoint.
  *
  * @param {object} attributes - Attributes fetched from manifest.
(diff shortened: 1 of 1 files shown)


scripts/editor/output-css-variables.js
```diff
@@ -231,6 +243,72 @@ export const outputCssVariables = (attributes, manifest, unique, globalManifest)
...
+	})
+}
+
+/**
+ * Iterating through variables matching the keys from responsiveAttributes and translating it to responsive attributes names.
+ *
+ * @param {object} responsiveAttributes - Responsive attributes that are read from component's/block's manifest.
+ * @param {object} variables            - Object containing objects with component's/block's attribute variables that are read from manifest.
+ *
+ * @return {object} Object prepared for setting all the variables to it's breakpoints.
```
goranalkovic-infinum, line 274:
```suggestion
 * @return {object} Object prepared for setting all the variables to its breakpoints.
```

goranalkovic-infinum (dismissed):
🎉 🎉 🎉

scripts/editor/output-css-variables.js
```diff
@@ -231,6 +243,72 @@ export const outputCssVariables = (attributes, manifest, unique, globalManifest)
 	return <style dangerouslySetInnerHTML={{__html: `${output} ${finalManualOutput}`}}></style>;
 }
 
+/**
+ * Sets up a breakpoint value to responsive attribute objects from responsiveAttribute object.
+ *
+ * @param {array}  attributeVariables  - Array of attribute variables object.
+ * @param {string} breakpointName    	 - Breakpoint name from responsiveAttribute's breakpoint in block's/component's manifest.
+ * @param {number} breakpointIndex   	 - Index of responsiveAttribute's breakpoint in manifest.
+ * @param {object} numberOfBreakpoints - Number of responsiveAttribute breakpoints in block's/component's manifest.
```
volfkarlo, line 252:
```suggestion
 * @param {number} numberOfBreakpoints - Number of responsiveAttribute breakpoints in block's/component's manifest.
```

iruzevic (changes requested):
looking good please add a little more detailed explanation what is happening in each line. tnx

scripts/editor/output-css-variables.js
```diff
@@ -231,6 +243,72 @@ export const outputCssVariables = (attributes, manifest, unique, globalManifest)
...
+		if (!responsiveAttributeName || _.isEmpty(variables[responsiveAttributeName])) {
+			return responsiveAttributesVariables;
+		}
+
+		// To be able to determine default breakpoint.
+		const numberOfBreakpoints = Object.entries(responsiveAttributeObject).length;
+		const responsiveAttributeVariables = Object.entries(responsiveAttributeObject).reduce((responsiveAttribute, [breakpointName, breakpointVariableName], breakpointIndex) => {
+			let breakpointVariables = {};
+
+			// Array treatment goes directly.
```
IvanGrginovInf, line 287: Could you elaborate on the comment? I.e. write it out in more details 😄

scripts/editor/output-css-variables.js
```diff
@@ -231,6 +243,72 @@ export const outputCssVariables = (attributes, manifest, unique, globalManifest)
...
+			if (Array.isArray(variables[responsiveAttributeName])) {
+				breakpointVariables = setBreakpointResponsiveVariables(variables[responsiveAttributeName], breakpointName, breakpointIndex, numberOfBreakpoints);
+
+				// Object treatment goes depending on a value inserted.
```
IvanGrginovInf, line 291: Same as above. Also check the indenting of this line.

IvanGrginovInf (changes requested):
Left a small comment regarding the comments, otherwise looks great, good job!

volfkarlo:
I've reordered and extracted the helper function outside so it looks as much as possible like the PHP part of the code.

No need to check it for now since I'll go through it once again with Adrian from JS to ensure everything is ok 🙂

scripts/editor/output-css-variables.js
```diff
@@ -72,6 +72,142 @@ export const globalInner = (itemValues, itemKey) => {
...
+
+				// Iterate options/multiple choices/boolean...
+				breakpointVariables = Object.entries(variables[responsiveAttributeName])
+				.reduce((acc, [attributeValue, attributeObject]) => {
+					return {
+						...acc,
+						[attributeValue]: setBreakpointResponsiveVariables(attributeObject, breakpointName, breakpointIndex, numberOfBreakpoints),
+					}
+				}, {});
+			}
```
dingo-d, line 138: This seems like it could be inverted and else could be removed (unnecessary else)

scripts/editor/output-css-variables.js
```diff
@@ -72,6 +72,142 @@ export const globalInner = (itemValues, itemKey) => {
...
+
+		// Set internal breakpoints variable.
+		let internalBreakpoints = [];
+
+		// If type default or value.
+		if (!Array.isArray(variableValue)) {
+			internalBreakpoints = variableValue[attributeValue] ?? [];
+		} else {
+			internalBreakpoints = variableValue;
+		}
```
dingo-d, line 176: Unnecessary `else`. You could just set the `internalBreakpoints` to `variableValue` and if the `variableValue` is an array, set it to the value inside the `if`.

dingo-d (changes requested):
Left some comments.

scripts/editor/output-css-variables.js
volfkarlo: Would `const internalBreakpoints =  Array.isArray(variableValue) ? variableValue : (variableValue[attributeValue] ?? []);`
be a better solution?
(to be honest, I didn't want to change it I just extracted the function)

I don't like that we are overriding variables at any point because any kind of modification might influence the outcome. And in this case, there is no need to make it anything other than constant. Would you agree?

scripts/editor/output-css-variables.js
volfkarlo: I don't see how else could be removed.

If I don't have an object I can't do `Object.entries(...)`. If I don't have an array, `setBreakpointResponsiveVariables` would receive an object instead of an array, and the function would break on that part.

volfkarlo: I mean, I can do something like this, but I don't see the point in writing two of the exact same returns just to avoid one else.

```
		const responsiveAttributeVariables = Object.entries(responsiveAttributeObject).reduce((responsiveAttribute, [breakpointName, breakpointVariableName], breakpointIndex) => {
			let breakpointVariables = {};

			if (Array.isArray(variables[responsiveAttributeName])) { // Array represents direct value(default or value).
				breakpointVariables = setBreakpointResponsiveVariables(variables[responsiveAttributeName], breakpointName, breakpointIndex, numberOfBreakpoints);
				return {
					...responsiveAttribute,
					[breakpointVariableName]: breakpointVariables,
				};
			}

			// Object treatment goes depending on a value inserted(multiple choice, boolean or similar).
			// Iterate options/multiple choices/boolean...
			breakpointVariables = Object.entries(variables[responsiveAttributeName])
			.reduce((acc, [attributeValue, attributeObject]) => {
				return {
					...acc,
					[attributeValue]: setBreakpointResponsiveVariables(attributeObject, breakpointName, breakpointIndex, numberOfBreakpoints),
				}
			}, {});

			// Collect all the values from one responsive attribute to one object.
			return {
				...responsiveAttribute,
				[breakpointVariableName]: breakpointVariables,
			};
		}, {});
```

But if that's a better practice, or it improves readability, it's ok with me.

scripts/editor/output-css-variables.js
dingo-d: What I meant is something like this:

```js
let internalBreakpoints = variableValue;

// If type default or value.
if (!Array.isArray(variableValue)) {
	internalBreakpoints = variableValue[attributeValue] ?? [];
}
```

That way the `internalBreakpoints` variable will always be set to something, but overwritten to `variableValue[attributeValue] ?? [];` if it's not an array.

> I don't like that we are overriding variables at any point because any kind of modification might influence the outcome. And in this case, there is no need to make it anything other than constant. Would you agree?

Not sure I understand the first part. You are already overriding the `internalBreakpoints` (you've initialized it to be an empty array).

volfkarlo: > Not sure I understand the first part. You are already overriding the internalBreakpoints (you've initialized it to be an empty array).

That's why I suggested
` const internalBreakpoints = Array.isArray(variableValue) ? variableValue : (variableValue[attributeValue] || []);`

The current
```
if (!Array.isArray(variableValue)) {
			internalBreakpoints = variableValue[attributeValue] ?? [];
		} else {
			internalBreakpoints = variableValue;
		}
```
was not written by me. This whole function was just extracted from the code so I can reuse it.

scripts/editor/output-css-variables.js
```diff
@@ -118,61 +255,22 @@ export const outputCssVariables = (attributes, manifest, unique, globalManifest)
...
-			// Set internal breakpoints variable.
-			let internalBreakpoints = [];
-
-			// If type default or value.
-			if (!Array.isArray(variableValue)) {
-				internalBreakpoints = variableValue[attributeValue] ?? [];
-			} else {
-				internalBreakpoints = variableValue;
-			}
-
```
volfkarlo, line 139: Here is the old code that I extracted.
@dingo-d

scripts/editor/output-css-variables.js
goranalkovic-infinum: Karlo's last suggestion is fine, although I definitely find Denis' version cleaner (initialized to value + if) to read.

BTW Karlo if you go with your last suggested ternary, make sure to change the _or_ operator to _nullish coalescing_
```js
const internalBreakpoints = 
    Array.isArray(variableValue) 
    ? variableValue 
    : (variableValue[attributeValue] || []);
```
to
```js
const internalBreakpoints = 
    Array.isArray(variableValue) 
    ? variableValue 
    : (variableValue[attributeValue] ?? []);
```

scripts/editor/output-css-variables.js
dingo-d: Sure leave it as is, it's super complicated to read, that's what was worrying me the most. Just good practices than anything.

scripts/editor/output-css-variables.js
dingo-d: You can even refactor to what Goc suggested, looks easier to read. Althougth I leave the ternary characters in the same line

```js
const internalBreakpoints = 
    Array.isArray(variableValue) ?
    variableValue :
    (variableValue[attributeValue] ?? []);
```

You decide which one looks best for you @volfkarlo 😄

scripts/editor/output-css-variables.js
goranalkovic-infinum: @dingo-d totally agree, but as returning null would change the output of the `.reduce()`, it would be a step back. Also it definitely is a bit hard to read, but I'm sure 3 reduces are still just as easy to read as the mess that would replace them would be.

volfkarlo: I shortened it and added 2 of the same return sentences 🙂

goranalkovic-infinum (approved):

dingo-d (approved):

iruzevic (approved):

iruzevic:
The issues were fixed, merged on request from Karlo due to Grginovs absence

PR #363, opened 2021-06-07
Props helper attribute renaming fix

goranalkovic-infinum:
The `props` helper was manifesting some weird behaviour with renaming some attribute names in nested components (for example (component1 -> component2 -> component3), so I switched it to use the safer `.replace()` method, which prevents slicing and dicing component names.


iruzevic:
please make the same fix on the PHP side as well

iruzevic (approved):

dingo-d (approved):

IvanGrginovInf (approved):
Approved but we definitely need to setup unit tests for this.

PR #379, opened 2021-07-01
Setup tests

IvanGrginovInf:
Setup JS tests for the following (currently will fail because props changes haven't been merged):
* Block registration and component attribute inheritance
* Props attribute generation on each lvl
* Both of these tests will automatically test all blocks defined in `src/init/Blocks` (actual blocks we have implemented) as well as `tests/data/src/Blocks` (mock block manifests we can use for testing some weird inheritance situations)

Block registration test
---
* We can manually define `expected` and `notExpected` attributes in each block inside `tests/data/block-attributes.js`

Props registration test
---
* We can manually define which attributes are expected / notExpected to be present in each lvl of block / component dependency tree as well as the new `prefix` attribute which is used to determine the `prefix` of the attribute we're settting in any lvl / component. This is done inside `tests/data/props-output` (will look if this can be automatically checked rather than having this pretty sizable object in there which is kind atedious to write for each component / block but this way at least we can manually set expectations for each lvl or we'll have to write tests for tests :D)

### DIFF
--- a/scripts/editor/icons/icons.js
+++ b/scripts/editor/icons/icons.js
@@ -1,4 +1,4 @@
-// import React from 'react';
+import React from 'react';
 
 /**
  * UI icons.
--- a/scripts/editor/icons/icons.js
+++ b/scripts/editor/icons/icons.js
@@ -1,4 +1,4 @@
-import React from 'react';
+// import React from 'react';
 
 /**
  * UI icons.
--- a/tests/helpers/blocks.js
+++ b/tests/helpers/blocks.js
@@ -81,18 +81,20 @@ export const recursiveBuildProps = (attributes, componentManifests, realName, ne
 	let propsArray = [];
 	const isNameDifferent = realName !== newName;
 	const newComponentAttributes = props(attributes, realName, isNameDifferent ? newName : '', isBlock);
-	propsArray = [...propsArray, {
+	const componentOutput = {
 		realName,
 		newName,
 		attributes: newComponentAttributes,
-	}];
-
+		subComponents: [],
+	};
+	
 	// Check if this component has any sub-components and recursively call this function.
-	const components = getComponentDependencies(componentManifests, realName);
+	const components = getComponentDependencies(componentManifests, realName, newName);
 	for (const subNewName of Object.keys(components)) {
 		const subRealName = components[subNewName];
-		propsArray = [...propsArray, ...recursiveBuildProps(newComponentAttributes, componentManifests, subRealName, subNewName)];
+		componentOutput.subComponents = [...componentOutput.subComponents, ...recursiveBuildProps(newComponentAttributes, componentManifests, subRealName, subNewName)]
 	}
-
+	
+	propsArray = [...propsArray, componentOutput];
 	return propsArray;
 }
--- a/tests/helpers/blocks.js
+++ b/tests/helpers/blocks.js
@@ -1,0 +1,52 @@
+import Path from 'path';
+import { readdirSync, readFileSync } from "fs";
+import { props } from '../../scripts/editor';
+
+const pathToBlocksFolder = Path.resolve(__dirname, '..', '..', 'blocks', 'init', 'src', 'Blocks');
+const pathToComponents = Path.resolve(pathToBlocksFolder, 'components');
+const pathToBlocks = Path.resolve(pathToBlocksFolder, 'custom');
+
+const pathToMockBlocksFolder = Path.resolve(__dirname, '..', 'data', 'src', 'Blocks');
+const pathToMockComponents = Path.resolve(pathToMockBlocksFolder, 'components');
+const pathToMockBlocks = Path.resolve(pathToMockBlocksFolder, 'custom');
+
+/**
+ * Returns all manifests from directories inside `dir`.
+ *
+ * @param {string} dir Full path to dir in which we're looking for other 
+ * 										 dirs (components / blocks) which have manifest.json inside.
+ * 										 Example: 'blocks/init/src/Blocks/components'
+ * @returns {array<object>} Array of manifest.json contents
+ */
+const getAllManifestsFromDir = (dir) => {
+	const manifestDirs = readdirSync(dir);
+	return manifestDirs.map((manifestDir) => {
+		const manifestPath = Path.resolve(dir, manifestDir, 'manifest.json');
+		const manifest = JSON.parse(readFileSync(manifestPath));
+		return manifest;
+	});
+}
+
+/**
+ * Get all components, both the mock ones specifically for tests and the actual libs blocks.
+ * 
+ * @return {array}
+ */
+export const getAllComponentManifests = () => {
+	const componentManifests = getAllManifestsFromDir(pathToComponents);
+	const mockComponentManifests = getAllManifestsFromDir(pathToMockComponents);
+	return [...componentManifests, ...mockComponentManifests];
+}
+
+/**
+ * Get all components, both the mock ones specifically for tests and the actual libs blocks.
+ *
+ * @return {array}
+ */
+export const getAllBlockManifests = () => {
+	const blockManifests = getAllManifestsFromDir(pathToBlocks);
+	const mockBlockManifests = getAllManifestsFromDir(pathToMockBlocks);
+	return [...blockManifests, ...mockBlockManifests];
+}
+
+/**
--- a/tests/helpers/blocks.js
+++ b/tests/helpers/blocks.js
@@ -50,3 +50,49 @@ export const getAllBlockManifests = () => {
 }
 
 /**
+ * Returns an object composed of all component dependencies (if they exist) if the following format:
+ * {
+ * 	 newName: realName
+ * }
+ *
+ * @param {array<object>} componentManifests Array of all component manifests
+ * @param {string} componentName Component name for which we want to get dependency components
+ * @return {object}
+ */
+export const getComponentDependencies = (componentManifests, componentName) => {
+	let components = {};
+
+	for(const componentManifest of componentManifests) {
+		if (componentManifest.componentName === componentName && componentManifest.components) {
+			components = componentManifest.components;
+			break;
+		}
+	}
+
+	return components;
+}
+
+/**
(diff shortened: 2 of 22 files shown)


scripts/editor/icons/icons.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+// import React from 'react';
```
goranalkovic-infinum, line 1: Why is this commented out? UI icons are JSX SVGs

tests/unit/blocks/blockRegistration.test.js
```diff
@@ -0,0 +1,174 @@
...
+		const attributeNames = Object.keys(getAttributes(globalManifest, wrapperManifest, componentManifests, blockManifest));
+
+		// Only test blocks which expected / not expected attributes defined.
+		if (!blockAttributes[blockName]) {
+			continue;
+		}
+
+		// Make sure the block contains all expected attributes
+		if (blockAttributes[blockName].expected) {
+			expect(
```
goranalkovic-infinum, line 114: Extra enter here?

goranalkovic-infinum (commented):
Left two questions, the rest looks good, nice job!

scripts/editor/icons/icons.js
IvanGrginovInf: Was testing out something when my tests were failing, forgot to return this, good catch.

tests/unit/blocks/blockRegistration.test.js
IvanGrginovInf: Not sure what you mean? This has the attributes spread over 2 lines so the message is more visible.

tests/helpers/blocks.js
```diff
@@ -0,0 +1,103 @@
...
+ *                     dirs (components / blocks) which have manifest.json inside.
+ *                     Example: 'blocks/init/src/Blocks/components'
+ * @returns {array<object>} Array of manifest.json contents
+ */
+const getAllManifestsFromDir = (dir) => {
+	const manifestDirs = readdirSync(dir);
+	return manifestDirs.map((manifestDir) => {
+		const manifestPath = Path.resolve(dir, manifestDir, 'manifest.json');
+		const manifest = JSON.parse(readFileSync(manifestPath));
+		return manifest;
```
dingo-d, line 26:
```suggestion
		return JSON.parse(readFileSync(manifestPath));
```

tests/helpers/blocks.js
```diff
@@ -0,0 +1,103 @@
...
+	for (const subNewName of Object.keys(components)) {
+		const subRealName = components[subNewName];
+		componentOutput.subComponents = [
+			...componentOutput.subComponents,
+			...recursiveBuildProps(newComponentAttributes, componentManifests, subRealName, subNewName)
+		];
+	}
+
+	propsArray = [...propsArray, componentOutput];
+	return propsArray;
```
dingo-d, line 102:
```suggestion
	return [...propsArray, componentOutput];
```

tests/unit/blocks/blockRegistration.test.js
```diff
@@ -0,0 +1,174 @@
...
+	// Repeat if expected has subComponents
+	if (expectedProps.components && Object.keys(expectedProps.components).length > 0) {
+		for (const componentKey of Object.keys(expectedProps.components)) {
+
+			// Expect built props has subComponents
+			expect(
+				buildProps,
+				`Missing "subComponents" key in built props for ${debugName(builtProps)}. Expecting one because this has subComponents defined in props-output.js`
+			).toHaveProperty('subComponents');
+			expect(
```
dingo-d, line 69:
```suggestion

			expect(
```

Just add space between these two, makes it more readable.

dingo-d (changes requested):
Left some comments. Nice 👍🏼

dingo-d (approved):

tests/unit/blocks/blockRegistration.test.js
goranalkovic-infinum: Looked a bit odd for me on the first look, that's why I asked. Maybe should've phrased it better 😄  It's all good now.

goranalkovic-infinum (approved):

tests/helpers/blocks.js
```diff
@@ -0,0 +1,125 @@
+import Path from 'path';
+import { readdirSync, readFileSync } from "fs";
```
volfkarlo, line 2: quite strange that they didn't put `readDirSync`

tests/helpers/blocks.js
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * Get all components, both the mock ones specifically for tests and the actual libs blocks.
+ * 
+ * @return {array}
+ */
+export const getAllComponentManifests = () => {
+	const componentManifests = getAllManifestsFromDir(pathToComponents);
+	const mockComponentManifests = getAllManifestsFromDir(pathToMockComponents);
+	return [...componentManifests, ...mockComponentManifests];
+}
```
volfkarlo, line 38: How come you didn't write it with

```suggestion
export const getAllComponentManifests = () => [...getAllManifestsFromDir(pathToComponents), ...getAllManifestsFromDir(pathToMockComponents)]
```

Not important at all, and maybe even less readable. But just a suggestion.

tests/helpers/blocks.js
```diff
@@ -0,0 +1,125 @@
...
+	const blockManifests = getAllManifestsFromDir(pathToBlocks);
+	const mockBlockManifests = getAllManifestsFromDir(pathToMockBlocks);
+	return [...blockManifests, ...mockBlockManifests];
+}
+
+/**
+ * Returns an object composed of all component dependencies (if they exist) if the following format:
+ * {
+ * 	 newName: realName
+ * }
```
volfkarlo, line 55:
```suggestion
 * {
 * 	 newName: realName,
 * 	 ...
 * }
```

Just a suggestion to clarify that there might be more of `newName: realName,` relations

tests/helpers/blocks.js
```diff
@@ -0,0 +1,125 @@
...
+ *
+ * @param {array<object>} componentManifests Array of all component manifests
+ * @param {string} componentName Component name for which we want to get dependency components
+ * @return {object}
```
volfkarlo, line 59:
```suggestion
 * @param {string} componentName Component name for which we want to get dependency components
 *
 * @return {object}
```

tests/helpers/blocks.js
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * Returns an object composed of all component dependencies (if they exist) if the following format:
+ * {
+ * 	 newName: realName
+ * }
+ *
+ * @param {array<object>} blockManifests Array of all block manifests
+ * @param {string} blockName Block name for which we want to get dependency components
+ * @return {object}
```
volfkarlo, line 82:
```suggestion
 * @param {string} blockName Block name for which we want to get dependency components
 *
 * @return {object}
```

tests/helpers/blocks.js
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	return components;
+}
+
+/**
+ * Returns an object composed of all component dependencies (if they exist) if the following format:
+ * {
+ * 	 newName: realName
+ * }
```
volfkarlo, line 78:
```suggestion
 * {
 * 	 newName: realName,
 * 	 ...
 * }
```

tests/unit/blocks/blockRegistration.test.js
```diff
@@ -0,0 +1,178 @@
...
+import { blockAttributes } from '../../data/block-attributes';
+import { expect } from '@jest/globals';
+import { getAllComponentManifests, getAllBlockManifests, getComponentDependencies, recursiveBuildProps, getBlockDependencies } from '../../helpers/blocks';
+import { propsOutput } from '../../data/props-output';
+
+/**
+ * Returns the name of the component from builtProps.
+ *
+ * @param {object} builtProps Built props for a component
+ * @returns {string}
```
volfkarlo, line 17:
```suggestion
 * @param {object} builtProps Built props for a component
 *
 * @returns {string}
```

tests/unit/blocks/blockRegistration.test.js
```diff
@@ -0,0 +1,178 @@
...
+	for (const blockManifest of blockManifests) {
+		const blockName = blockManifest.blockName;
+		const attributeNames = Object.keys(getAttributes(globalManifest, wrapperManifest, componentManifests, blockManifest));
+
+		// Only test blocks which expected / not expected attributes defined.
+		if (!blockAttributes[blockName]) {
+			continue;
+		}
+
+		// Make sure the block contains all expected attributes
```
volfkarlo, line 116:
```suggestion
		// Make sure the block contains all expected attributes.
```

tests/unit/blocks/blockRegistration.test.js
```diff
@@ -0,0 +1,178 @@
...
+		if (blockAttributes[blockName].expected) {
+			expect(
+				attributeNames,
+				`Block ${blockName} does not contain all expected attributes`
+			).toEqual(expect.arrayContaining(blockAttributes[blockName].expected));
+		}
+
+		// Make sure the block does not contain nonExpected attributes
```
volfkarlo, line 124:
```suggestion
		// Make sure the block does not contain nonExpected attributes.
```

volfkarlo (changes requested):
👏 

Just a few new lines at the end of the files are missing, and minor comments. Other than that great job! 😄

gabriel-glo (approved):

volfkarlo (commented):
I've changed form Request changes to comment because these are all just cosmetic changes.

volfkarlo (approved):
Sorry for notifications, it didn't change(from request changes to commnet) so I changed to Approve

tests/helpers/blocks.js
IvanGrginovInf: Generally I like keeping things in variables in case you need to console.log it in the future. But I understand it's a bit worse, will change this in next PR in which I'm fixing tests. 👍

PR #378, opened 2021-06-30
updating get attributes method for nesting

iruzevic:
- updating get_attributes method for prop drilling and prefixing attributes

PR #380, opened 2021-07-01
refactoring props helper for new attribute prefixes way

iruzevic:
- updating props helper for better drilling in combination with the previous PR #378
- updating checkAttr helper to use parent prefix if it is used.
- making a new helper for saving values in the setAttributes function

### DIFF
--- a/scripts/helpers/check-attr.js
+++ b/scripts/helpers/check-attr.js
@@ -6,7 +6,7 @@ import _ from 'lodash';
  *
  * @param {string} key                       - Key to check.
  * @param {array} attributes                 - Array of attributes.
- * @param {object} manifest                  - Default attributes from manifest.json.
+ * @param {object} manifest                  - Components/blocks manifest.json
  * @param {boolean} [undefinedAllowed=false] - Allowed detection of undefined values.
  *
  * @return {mixed}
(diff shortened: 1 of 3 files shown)


goranalkovic-infinum (approved):

scripts/helpers/check-attr.js
```diff
@@ -12,8 +13,21 @@ import _ from 'lodash';
...
 
-	if (Object.prototype.hasOwnProperty.call(attributes, key)) {
-		return attributes[key];
+	// Check if there is prefix in the attributes object.
+	const prefix = attributes?.prefix;
+	let newKey = key;
+
+	// If there is no prefix return the key as it was.
+	// If there is a prefix, remove the attribute component name prefix and replace it with the new prefix.
+	if (typeof prefix !== 'undefined') {
```
goranalkovic-infinum, line 22:
```suggestion
	if (prefix) {
```

By using `?.` you're sure that it will never be undeclared, just that it can be `undefined`, so no need for `typeof` checks.

goranalkovic-infinum (approved):
Added 1 suggestion

scripts/helpers/check-attr.js
iruzevic: I have tested this and it is not working, sorry

goranalkovic-infinum (approved):

dingo-d (approved):

IvanGrginovInf (approved):

PR #381, opened 2021-07-02
updating blocks/component to incorporate new helpers

iruzevic:
- refactoring blocks and components for new changes on props helper and fixes on new getAttrKey helper.
- updating image component to remove ID from the setAttribute option.
- this is an repetitive work for all items

### DIFF
--- a/blocks/init/src/Blocks/components/accordion/components/accordion-editor.js
+++ b/blocks/init/src/Blocks/components/accordion/components/accordion-editor.js
@@ -2,19 +2,17 @@ import React from 'react';
 import { __ } from '@wordpress/i18n';
 import classnames from 'classnames';
 import { RichText } from '@wordpress/block-editor';
-import { selector, checkAttr } from '@eightshift/frontend-libs/scripts/helpers';
+import { selector, checkAttr, getAttrKey } from '@eightshift/frontend-libs/scripts/helpers';
 import manifest from './../manifest.json';
 
 export const AccordionEditor = (attributes) => {
 	const {
-		componentName: manifestComponentName,
 		componentClass: manifestComponentClass,
 		options: manifestOptions,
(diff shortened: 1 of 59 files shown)


blocks/init/src/Blocks/components/accordion/components/accordion-editor.js
```diff
@@ -43,7 +40,7 @@ export const AccordionEditor = (attributes) => {
 						<RichText
 							placeholder={placeholder}
 							value={accordionTitle}
-							onChange={(value) => setAttributes({ [`${componentName}Title`]: value })}
+							onChange={(value) => setAttributes({ [getAttrKey('accordionTitle', attributes, manifest)]: value })}
```
dingo-d, line 43: How come you didn't use the `${componentName}Title` for the first argument in the `getAttrKey`?

iruzevic: in general we can use it but we didn't use it in the checkAttr and this has the same props as checkAttr  and it is easier this way for the user to understand what is happening because the full attribute name corresponds to the name defined in the manifest under the attributes key. For me this looks cleaner

blocks/init/src/Blocks/components/card/components/card-options.js
```diff
@@ -22,7 +22,7 @@ export const CardOptions = (attributes) => {
 			<hr />
 
 			<HeadingOptions
-				{...props(attributes, 'heading', 'intro')}
+				{...props(attributes, 'intro')}
```
goranalkovic-infinum, line 25: This looks awesome! 😄

goranalkovic-infinum (approved):

iruzevic:
I will merge this because this is a simple PR with nothing major here

PR #402, opened 2021-07-20
Goc's goodie basket

goranalkovic-infinum:
After a bunch of work on some projects, I decided to bunch up all the goodies and bring them to Frontend libs - to everyone!

# Changelog

### Block icons
1 new Block icon (`es-prev-next`) and 1 update (`es-code`).

image

### UI icons
Added **53 new icons** and updated a couple of existing ones.

Here are all icons we have currently:
image

### LinkToolbarButton
Updated to work with `props` and non-hardcoded attribute names.

**Breaking change** (minor as the component wasn't used widely)
- `componentName`  attribute isn't used anymore
- added new option for deciding if new tab options is displayed - `showNewTabOption`
- added 2 new attributes: `urlAttrName` and `isNewTabAttrName` (optional) for providing the correct attribute names to be set - should be used with `getAttrKey`

### Other
- updated FrontendLibs button to adapt new changes

### LinkEditComponent
Provides an inline link picker, made for the Options panel, and to match all the current Gutenberg styling.

image
image

### ColorPickerComponent
Provides a compact, inline color picker for the Options panel, to reduce clutter in components/blocks where many color are used.

image
image

### InlineNotification
A simple inline notification, a more compact replacement for Gutenberg's `Notice` component.

- `warning`, `error` and `info` types (select them via the `InlineNotificationType` enum from the same file)
- option for high-visibility outline to make sure it's visible on any background

image

### ComponentUseToggle
Adds a fancy replacement for label + classic component use toggle, which can adapt to show only label if needed without taking too much space.

image
image

### IconToggle
A right-aligned ToggleControl with an icon on the left, to better fit in with the rest of the components.

image

### SimpleVerticalSingleSelect
A nice way to provide selection for a couple of options (best used for 2 to 5-6 options).

image

### truncateMiddle
A simple helper that truncates a word by cutting it in the middle.
Max length and truncation character are configurable.


IvanGrginovInf (approved):

dingo-d (approved):
🤯 #madRespect

scripts/components/color-picker-component/docs/readme.mdx
```diff
@@ -0,0 +1,33 @@
...
+1. Import the component from Frontend Libs
+2. Add the `LinkEditComponent` into the toolbar
+
+```jsx
+import { LinkEditComponent } from '@eightshift/frontend-libs/scripts/components';
+import { getAttrKey } from '@eightshift/frontend-libs/scripts/helpers';
+
+// ...
+
+export const ButtonOptions = (attributes) =>  {
```
iruzevic, line 17: wrong example here

iruzevic (changes requested):

scripts/components/color-picker-component/docs/readme.mdx
goranalkovic-infinum: nice catch, thanks!

dingo-d (dismissed):

iruzevic (approved):

scripts/components/icon-toggle/docs/story.js
```diff
@@ -0,0 +1,38 @@
...
+			/>
+
+			<br />
+
+			<IconToggle
+				icon={icons.width}
+				label='Full width'
+				checked={current}
+				onChange={(value) => setCurrent(value)}
+				help='Help text here.'
```
dingo-d, line 34: Not translatable?

dingo-d (changes requested):
Left a comment

scripts/components/color-picker-component/color-picker-component.js
```diff
@@ -0,0 +1,96 @@
...
+ * @param {React.Component?} props.label                        - Label to represent the control
+ * @param {string?} [props.textDomain=eightshift-frontend-libs] - Text domain to use for i18n.
+ * @param {boolean} [props.canReset=true]                       - If `true`, a clear/reset button is shown.
+ */
+export const ColorPickerComponent = ({
+	colors,
+	value,
+	onChange,
+	label,
+	textDomain = 'eightshift-frontend-libs',
```
volfkarlo, line 24: Is it a real use case scenario to send text-domain in `ColorPickerComponent`? I know that when you build up the `eightshift-frontend-libs` it switched depending on the project(for example when you use `eightshift-frontend-libs` in redesign web it is all searched and replaced with local textdomain - not 100% sure but I believe it is like that)

scripts/components/icon-toggle/docs/story.js
volfkarlo: This question applies to line 22, 31, and 34 😄

scripts/components/inline-notification/docs/readme.mdx
```diff
@@ -0,0 +1,13 @@
+# InlineNotification
+
+A simple component to show inline editor notifications (most often for very simple messages and validation).
+
+## Usage
+
+```jsx
+<InlineNotification
+	text='This is an error!'
```
volfkarlo, line 9: Also translatable(even though it's just an example).

scripts/components/inline-notification/docs/story.js
```diff
@@ -0,0 +1,54 @@
...
+};
+export const regular = () => {
+	return (
+		<div style={{
+			display: 'flex',
+			flexDirection: 'column',
+			gap: '1rem',
+		}}>
+			<InlineNotification
+				text='This is a warning!'
```
volfkarlo, line 21: Translations.

scripts/components/inline-notification/inline-notification.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import { icons } from '@eightshift/frontend-libs/scripts/editor';
+import classnames from 'classnames';
+
+export const InlineNotificationType = {
+	warning: 'warning',
+	error: 'error',
+	info: 'info'
+}
```
volfkarlo, line 9: I like this enum solution a lot(at least it reminds me of it), awesome stuff. If this is the case, I believe types should be in CAPS.

```suggestion
export const InlineNotificationType = {
	WARNING: 'warning',
	ERROR: 'error',
	INFO: 'info'
}
```

volfkarlo: Also I believe since you are able to export it, you can add docs of each type.

```suggestion
/**
* ERROR - when something will fall apart
* WARNING - not falling apart but not good
* ...
*/
export const InlineNotificationType = {
	WARNING: 'warning',
	ERROR: 'error',
	INFO: 'info'
}
```

scripts/components/inline-notification/inline-notification.js
```diff
@@ -0,0 +1,48 @@
...
+		case InlineNotificationType.error:
+			icon = icons.errorCircle;
+			break;
+		case InlineNotificationType.warning:
+			icon = icons.warning;
+			break;
+		default:
+			icon = icons.infoCircle;
+			break;
+	}
```
volfkarlo, line 36: you could maybe make another object something like

```suggestion
	type = InlineNotificationType.warning,
	showContrastOutline = false,
}) => {
	const notificationIcon = {
		warning: 'warning'
		error: 'errorCircle',
		somethingElse: 'redTurboError',
	}
```
and then call
```suggestion
	icon = notificationIcon[type] || notification.infoCircle;
```

scripts/components/link-edit-component/link-edit-component.js
```diff
@@ -0,0 +1,148 @@
...
+		if (showNewTabOption) {
+			newValues = {
+				...newValues,
+				[isNewTabAttrName]: undefined,
+			};
+		}
+
+		setAttributes(newValues);
+
+		setIsDropdownOpen(false);
```
volfkarlo, line 55: Are you sure it will come to this point?

I think that after`setAttributes` code there comes a new state and rerender is coming. Once I had problems with it that's why I'm asking.

scripts/components/link-edit-component/link-edit-component.js
```diff
@@ -0,0 +1,148 @@
...
+		urlSettings = [...urlSettings, {
+			id: 'opensInNewTab',
+			title: __('Open in new tab', textDomain),
+		}
+		];
+	}
+
+	let currentValue = {
+		url: url
+	};
```
volfkarlo, line 70:
```suggestion
	let currentValue = {
		url,
	};
```

scripts/components/link-edit-component/link-edit-component.js
```diff
@@ -0,0 +1,148 @@
...
+
+	if (showNewTabOption) {
+		currentValue = {
+			...currentValue,
+			opensInNewTab: opensInNewTab,
+		}
```
volfkarlo, line 76:
```suggestion
		currentValue = {
			...currentValue,
			opensInNewTab,
		}
```

scripts/components/link-toolbar-button/link-toolbar-button.js
```diff
@@ -9,72 +9,105 @@ import { __, sprintf } from '@wordpress/i18n';
...
+		urlSettings = [...urlSettings, {
+			id: 'opensInNewTab',
+			title: __('Open in new tab', textDomain),
+		}
+		];
+	}
+
+	let currentValue = {
+		url: url
+	};
```
volfkarlo, line 69:
```suggestion
	let currentValue = {
		url,
	};
```

scripts/components/link-toolbar-button/link-toolbar-button.js
```diff
@@ -9,72 +9,105 @@ import { __, sprintf } from '@wordpress/i18n';
...
+
+	if (showNewTabOption) {
+		currentValue = {
+			...currentValue,
+			opensInNewTab: opensInNewTab,
```
volfkarlo, line 74:
```suggestion
		currentValue = {
			...currentValue,
			opensInNewTab,
```

styles/override-editor.scss
```diff
@@ -614,4 +615,98 @@ body {
...
 	}
+
+	// Simple button row/toolbar.
+	.es-simple-editor-button-row {
+		display: flex;
+		gap: 0.5rem;
+		margin-bottom: 0.5rem;
+	}
+
+	// Decorative/unselectable text
```
volfkarlo, line 626:
```suggestion
	// Decorative/unselectable text.
```

styles/override-editor.scss
```diff
@@ -614,4 +615,98 @@ body {
...
+	.es-decorative-text {
+		opacity: 0.6;
+		user-select: none;
+	}
+
+	// ColorPickerComponent inner color picker container
```
volfkarlo, line 632:
```suggestion
	// ColorPickerComponent inner color picker container.
```

styles/override-editor.scss
```diff
@@ -614,4 +615,98 @@ body {
...
+	.es-color-picker-component__inner-container {
+		padding: 0.5rem 1rem;
+	}
+
+	// ColorPickerComponent current color button icon
```
volfkarlo, line 637:
```suggestion
	// ColorPickerComponent current color button icon.
```

volfkarlo (changes requested):
Few micro comments. Awesome stuff Goc! You are our HERO! 🎉 🦸

scripts/components/icon-toggle/docs/story.js
goranalkovic-infinum: It's just a storybook example, while it can be translated, didn't see the need to do it here as I found thee translations to be a bit wonky in storybook. If you think they should be there, I'll add them 😄

scripts/components/link-edit-component/link-edit-component.js
goranalkovic-infinum: https://user-images.githubusercontent.com/77000136/126447332-fdf12096-70ce-4150-8aba-18b350fcc9e6.mov

Yup, works as intended. My guess is because it's handling a popover and a seperate state, the re-render of the parent block is not affecting it.

scripts/components/color-picker-component/color-picker-component.js
goranalkovic-infinum: This lives in `node_modules`, so I doubt any replacements get made, but if they do, I can very easily remove the `textDomain` and make it even simpler 😄 

At the end of the day, it's an optional parameter with a default value, so nothing will break if you forget to pass it.

scripts/components/inline-notification/inline-notification.js
goranalkovic-infinum: Made it a switch here in case I decide to add something in the future - with a switch it would be less hassle to add/change a new variable than with an object.

scripts/components/color-picker-component/color-picker-component.js
iruzevic: in my opinion it would be better to pass the full string from the project and not the textDomain so that the project wories about translation

scripts/components/inline-notification/docs/story.js
iruzevic: storybook doesn't need translations it is dummy data

scripts/components/icon-toggle/docs/story.js
goranalkovic-infinum: I've checked out other storybook examples, none of them use localization, just straight-up strings. 🤷

iruzevic: storybook doesn't need translations it is dummy data

goranalkovic-infinum: Added them 😄

scripts/components/color-picker-component/color-picker-component.js
goranalkovic-infinum: That would add probably 4-6 attributes to every component (for every string), I'm worried it would get a bit redundant and bloated if it was like this:

```jsx
<ColorPickerComponent
	label={
		<>
			{icons.color}
			<span>My color picker</span>
		</>
	}
	colors={getPaletteColors()}
	value={objData.color}
	onChange={(value) => setObjData({ color: value })}
        pickColorLabel={__('Pick a color', 'eightshift-frontend-libs')}
        buttonNoValueLabel={__('Change color', 'eightshift-frontend-libs')}
        buttonEditLabel={__('Pick a color', 'eightshift-frontend-libs')}
        resetButtonLabel={__('Reset', 'eightshift-frontend-libs')}
/>
```

iruzevic: well if you have defaults on the props there is no harm in having them like this. in 99% cases you will not set any of them so ho cares if we have 2 od 20 props that you can pass. :D

goranalkovic-infinum: Wouldn't the defaults have the non-project text domain, though? Or does it get auto-replaced somehow? 😄

iruzevic: If it goes to you project it will get replaced if not it will stay the same. Lets not waste time on this check how other core components have this and make it that way ;)

goranalkovic-infinum: Yes, sir!

goranalkovic-infinum: Looks like core is not using the second argument at all, I'll implement it like that.

iruzevic (approved):

scripts/components/link-toolbar-button/link-toolbar-button.js
```diff
@@ -8,76 +8,109 @@ import { __, sprintf } from '@wordpress/i18n';
...
+	if (showNewTabOption) {
+		urlSettings = [...urlSettings, {
+			id: 'opensInNewTab',
+			title: newTabOptionName,
+		}
+		];
+	}
+
+	let currentValue = {
+		url
```
volfkarlo, line 69:
```suggestion
		url,
```

It doesn't matter

volfkarlo (approved):

dingo-d (approved):

PR #427, opened 2021-08-24
Add modal component

simun-ivanac:
Simple modal window which displays passed content.
It uses [https://micromodal.vercel.app/](https://micromodal.vercel.app/) library.

It’s still work in progress :)


blocks/init/src/Blocks/components/modal/assets/index.js
```diff
@@ -0,0 +1,24 @@
+import domReady from '@wordpress/dom-ready';
+import manifest from './../manifest.json';
+import MicroModal from 'micromodal';
+
+domReady(() => {
```
iruzevic, line 5: for all frontend js you must use dynamic import. please red more on this link:

https://infinum.github.io/eightshift-docs/docs/basics/dynamic-import

blocks/init/src/Blocks/components/modal/assets/index.js
```diff
@@ -0,0 +1,24 @@
...
+	const elements = document.querySelectorAll(selector);
+	const body = document.getElementsByTagName('body')[0];
+
+	// Append all modals at the bottom of body.
+	[...elements].forEach((element) => {
+		body.append(element);
+	});
+
+	// Initialize MicroModal.
+	MicroModal.init({
```
iruzevic, line 16: please use js selectors to open and close modal just so If you want to change the library later on the selectors can stay the same.

blocks/init/src/Blocks/components/modal/components/modal-editor.js
```diff
@@ -0,0 +1,64 @@
...
+	const modalExitButtonClass = selector(componentClass, componentClass, 'close-button');
+
+	return (
+		<div className={modalClass} data-id={unique} id={modalId} aria-hidden="false">
+			{outputCssVariables(attributes, manifest, unique, globalManifest)}
+
+			<div className={modalOverlayClass} tabIndex="-1" data-micromodal-close>
+				<div className={modalDialogClass} role="dialog" aria-modal="true">
+					{modalExitButton &&
+						<button
```
iruzevic, line 49: why not use button component that you have in the library?

blocks/init/src/Blocks/components/modal/components/modal-options.js
```diff
@@ -0,0 +1,48 @@
...
+		<>
+			<ComponentUseToggle
+				label={label}
+				checked={modalUse}
+				onChange={(value) => setAttributes({ [getAttrKey('modalUse', attributes, manifest)]: value })}
+				showUseToggle={showModalUse}
+				showLabel={showLabel}
+			/>
+
+			{modalUse &&
```
iruzevic, line 33: please use "modalUse like in the modal-editor.js" file

blocks/init/src/Blocks/components/modal/modal-style.scss
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	&__overlay {
+		position: fixed;
+		top: 0;
+		left: 0;
+		right: 0;
+		bottom: 0;
+		z-index: 999;
+		background-color: rgba(0, 0, 0, 0.5);
```
iruzevic, line 19: never use colors direct like this. you can use only what you have in the global manifest.json

https://infinum.github.io/eightshift-docs/docs/basics/blocks-styles#global-variables

blocks/init/src/Blocks/components/modal/modal.php
```diff
@@ -0,0 +1,85 @@
...
+?>
+
+<div
+	class="<?php echo \esc_attr($modalClass); ?>"
+	data-id="<?php echo \esc_attr($unique); ?>"
+	id="<?php echo \esc_attr($modalId); ?>"
+	aria-hidden="true"
+>
+	<?php
+		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
iruzevic, line 53: ignores goes in the same line as the function after. please check the other components

blocks/init/src/Blocks/components/modal/modal.php
```diff
@@ -0,0 +1,85 @@
...
+					data-micromodal-close
+				>
+					<?php echo $manifest['resources']['icon']; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>
+				</button>
+			<?php } ?>
+
+			<div class="<?php echo \esc_attr($modalContentClass); ?>">
+				<?php
+					// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+					echo $modalContent; // @phpstan-ignore-line
```
iruzevic, line 80: ignore line in the same line as up

iruzevic (changes requested):
- All issues from js editor files applies to php file.
- please implement story

blocks/init/src/Blocks/components/modal/components/modal-editor.js
```diff
@@ -0,0 +1,64 @@
...
+	]);
+
+	const modalOverlayClass = selector(componentClass, componentClass, 'overlay');
+	const modalDialogClass = selector(componentClass, componentClass, 'dialog');
+	const modalContentClass = selector(componentClass, componentClass, 'content');
+	const modalExitButtonClass = selector(componentClass, componentClass, 'close-button');
+
+	return (
+		<div className={modalClass} data-id={unique} id={modalId} aria-hidden="false">
+			{outputCssVariables(attributes, manifest, unique, globalManifest)}
```
mbmjertan, line 44: You're not using CSS variables in this block, so this inspires a question for the whole team - should we give data-ids and use `outputCssVariables` when bootstrapping blocks that don't use CSS variables at that point? (for future-proofing)

Or should we add CSS variable support to blocks only when they use them? Both approaches make sense, but seems that we should be consistent about this.

blocks/init/src/Blocks/components/modal/components/modal-editor.js
```diff
@@ -0,0 +1,64 @@
...
+
+			<div className={modalOverlayClass} tabIndex="-1" data-micromodal-close>
+				<div className={modalDialogClass} role="dialog" aria-modal="true">
+					{modalExitButton &&
+						<button
+							className={modalExitButtonClass}
+							aria-label={__('Close modal', 'eightshift-frontend-libs')}
+							data-micromodal-close
```
mbmjertan, line 52: > @iruzevic:
> why not use button component that you have in the library?

Probably because of this data attribute. While it would introduce a (small) markup difference between frontend and the editor, I believe that this line should be left out. It doesn't actually do anything, as Micromodal can't (shouldn't) be loaded in the editor and it tightens this block's coupling with the library. If I'm not mistaken, there aren't any other caveats to using the button component here.

blocks/init/src/Blocks/components/modal/modal-style.scss
```diff
@@ -0,0 +1,88 @@
...
+		display: flex;
+		justify-content: center;
+		align-items: center;
+		will-change: transform;
```
mbmjertan, line 23: > Warning: will-change is intended to be used as a last resort, in order to try to deal with existing performance problems. It should not be used to anticipate performance problems.
> https://developer.mozilla.org/en-US/docs/Web/CSS/will-change

Are you having performance problems with your animations?

blocks/init/src/Blocks/components/modal/modal-style.scss
```diff
@@ -0,0 +1,88 @@
...
+
+	&[aria-hidden='true'] {
+		#{$this}__overlay {
+			animation: modalFadeOut 0.3s cubic-bezier(0, 0, 0.2, 1);
+		}
+	
+		#{$this}__dialog {
+			animation: modalSlideOut 0.3s cubic-bezier(0, 0, 0.2, 1);
+		}
+	}
```
mbmjertan, line 87: You should probably be able to use CSS transitions for this (depending on what you do with the `display` property)

blocks/init/src/Blocks/components/modal/modal.php
mbmjertan: Ignore statements count towards line length limits, so some components' PHP implementations (image, column, columns, ...) put them on the line above to avoid having another lint error. Seems okay to me :shrug:
(It applies to the next line if there isn't PHP code on the same line)

blocks/init/src/Blocks/components/modal/modal.php
```diff
@@ -0,0 +1,85 @@
...
+>
+	<?php
+		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+		echo Components::outputCssVariables($attributes, $manifest, $unique, $globalManifest);
+	?>
+
+	<div
+		class="<?php echo \esc_attr($modalOverlayClass); ?>"
+		tabIndex="-1"
+		data-micromodal-close
```
mbmjertan, line 60: Same comment regarding Micromodal coupling. If the library prevents you from doing things differently, that's okay, but generally a more library-agnostic approach is better.

mbmjertan (changes requested):
Good work, but a few comments / suggestions.

blocks/init/src/Blocks/components/modal/modal.php
iruzevic: i generally don't like for comments to interfere in reading the code but it is ok

blocks/init/src/Blocks/components/modal/assets/index.js
```diff
@@ -0,0 +1,24 @@
...
+
+domReady(() => {
+	const selector = `.${manifest.componentJsClass}`;
+	const elements = document.querySelectorAll(selector);
+	const body = document.getElementsByTagName('body')[0];
+
+	// Append all modals at the bottom of body.
+	[...elements].forEach((element) => {
+		body.append(element);
+	});
```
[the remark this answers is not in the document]
iruzevic: I told him to do this. This is because you can for example use modal to show video from the carousel and due to z-indexes and positions all can break. So if you transfer all modals to top level Dom you are ok

blocks/init/src/Blocks/components/modal/components/modal-editor.js
iruzevic: if you are not using it remove it

blocks/init/src/Blocks/components/modal/components/modal-editor.js
iruzevic: if this is just an X you can use it like this no problem. I wanted to se how Šimun thinks ;)

blocks/init/src/Blocks/components/modal/assets/index.js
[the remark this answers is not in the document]
simun-ivanac: It's basically to avoid styling issues and to act independently from block (even if they are part of if).

blocks/init/src/Blocks/components/modal/modal.php
simun-ivanac: it's not problem, I'll fix it 😃

blocks/init/src/Blocks/components/modal/components/modal-editor.js
simun-ivanac: This is my bad
I'll remove it

blocks/init/src/Blocks/components/modal/components/modal-editor.js
simun-ivanac: I'll change that as well
small difference wont make a big deal

blocks/init/src/Blocks/components/modal/modal.php
```diff
@@ -76,8 +71,7 @@ class="<?php echo \esc_attr($modalExitButtonClass); ?>"
 
 			<div class="<?php echo \esc_attr($modalContentClass); ?>">
 				<?php
-					// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
-					echo $modalContent; // @phpstan-ignore-line
+					echo $modalContent; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
IvanGrginovInf, line 74: These PHPStan errors can be ignored, fixed them in this PR: https://github.com/infinum/eightshift-libs-stubs/pull/4

blocks/init/src/Blocks/components/modal/modal-style.scss
```diff
@@ -0,0 +1,88 @@
...
+
+	@keyframes modalFadeOut {
+		from { opacity: 1; }
+		  to { opacity: 0; }
+	}
+
+	@keyframes modalSlideIn {
+	  from { transform: translateY(15%); }
+		to { transform: translateY(0); }
+	}
```
IvanGrginovInf, line 62: Please fix spacing here

blocks/init/src/Blocks/components/modal/modal-style.scss
```diff
@@ -0,0 +1,88 @@
...
+
+	&__dialog {
+		min-width: 360px;
+		max-width: calc(100% - 1.25rem);
+		max-height: 90%;
+		overflow: auto;
+		position: relative;
+		padding: 1.25rem;
+		will-change: transform;
```
IvanGrginovInf, line 33: No need to add that, that's a "last measure" performance optimization (https://developer.mozilla.org/en-US/docs/Web/CSS/will-change) Plus you shouldn't really have any lag issues with animating transforms (`opacity` and `transform` are generally the only things you should animate if possible).

blocks/init/src/Blocks/components/modal/modal.php
```diff
@@ -0,0 +1,79 @@
...
+			aria-modal="true"
+		>
+			<?php if ($modalExitButton) { ?>
+				<button
+					class="<?php echo \esc_attr($modalExitButtonClass); ?>"
+					aria-label="Close modal"
+					data-micromodal-close
+				>
+					<?php echo $manifest['resources']['icon']; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>
+				</button>
```
IvanGrginovInf, line 69: Here you should use `Button` component if possible. You can pass arbitrary data-attributes to it via the `buttonAttrs` attribute.

blocks/init/src/Blocks/components/modal/docs/readme.mdx
```diff
@@ -0,0 +1,11 @@
...
+
+Component used to provide modal functionality to your project.
+
+## Implementation
+
+Open a terminal and type in this command inside your projects root:
+
+```shell
+wp boilerplate use_component --name=modal
+```
```
IvanGrginovInf, line 11: If you use the button component from frontend libs (as you should), you can specify it here as a dependency (see the card's readme for example / structure)

blocks/init/src/Blocks/components/modal/modal.php
IvanGrginovInf: Yes, ideally you would have trigger data-attribute names as component attributes (and you can have them default to micromodal's values) and then just echo them here.

IvanGrginovInf (changes requested):
Left a couple of comments but otherwise looks great, gj!

blocks/init/src/Blocks/components/modal/modal-style.scss
simun-ivanac: I see what you mean and that’s true
It can’t be animated, but the library can handle it that way
But anyway, since library could be replaced, I’ll change that

blocks/init/src/Blocks/components/modal/components/modal-editor.js
```diff
@@ -0,0 +1,67 @@
...
+		<div className={modalClass} id={modalId} aria-hidden="false">
+			<div className={modalOverlayClass} tabIndex="-1">
+				<div className={modalDialogClass} role="dialog" aria-modal="true">
+					{modalExitButton &&
+						<div className={modalCloseClass}>
+							<button
+								className={modalCloseButtonClass}
+								aria-label={__('Close modal', 'eightshift-frontend-libs')}
+								dangerouslySetInnerHTML={{ __html: icon }}
+								onClick={onClick}
```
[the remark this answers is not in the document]
simun-ivanac: Modal needs to receive signal when to open, so I passed the function to handle closing (see story).
It’s probably temporary, and mostly part of implementation.

blocks/init/src/Blocks/components/modal/manifest.json
```diff
@@ -0,0 +1,34 @@
...
+		"attributes": {
+			"modalUse": true,
+			"modalExitButton": true,
+			"modalContent": "This is modal",
+			"modalId": "modal-id"
+		}
+	},
+	"attributes": {
+		"modalUse": {
+			"type": "boolen",
```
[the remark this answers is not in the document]
simun-ivanac: typo... fixed

blocks/init/src/Blocks/components/modal/manifest.json
```diff
@@ -0,0 +1,34 @@
...
+			"default": true
+		},
+		"modalExitButton": {
+			"type": "boolen",
```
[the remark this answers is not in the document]
simun-ivanac: typo... fixed
but it didn't throw any errors or created some issues... hm

IvanGrginovInf (changes requested):
Please fix the phpstan errors, no need to ignore `phpstan errors` if there's no error.

iruzevic (dismissed):
Looks good

blocks/init/src/Blocks/components/modal/modal.php
```diff
@@ -0,0 +1,75 @@
...
+?>
+
+<div
+	class="<?php echo \esc_attr($modalClass); ?>"
+	id="<?php echo \esc_attr($modalId); ?>"
+	aria-hidden="true"
+>
+	<div
+		<?php echo \esc_attr("data-{$componentJsToggleClass}-close"); ?>
+		class="<?php echo \esc_attr("{$componentClass}__overlay"); ?>"
```
mbmjertan, line 47: Use the `selector` helper please.

blocks/init/src/Blocks/components/modal/modal.php
```diff
@@ -0,0 +1,75 @@
...
+		tabIndex="-1"
+	>
+		<div
+			class="<?php echo \esc_attr("{$componentClass}__dialog"); ?>"
```
mbmjertan, line 51: Use the `selector` helper please.

blocks/init/src/Blocks/components/modal/modal.php
```diff
@@ -0,0 +1,75 @@
...
+			role="dialog"
+			aria-modal="true"
+		>
+			<?php if ($modalExitButton) { ?>
+				<div class="<?php echo \esc_attr("{$componentClass}__close"); ?>">
```
mbmjertan, line 56: Use the `selector` helper please.

blocks/init/src/Blocks/components/modal/modal.php
```diff
@@ -0,0 +1,75 @@
...
+					<button
+						<?php echo \esc_attr("data-{$componentJsToggleClass}-close"); ?>
+						class="<?php echo \esc_attr("{$componentClass}__close-button"); ?>"
```
mbmjertan, line 59: Use the `selector` helper please.

blocks/init/src/Blocks/components/modal/modal.php
```diff
@@ -0,0 +1,75 @@
...
+						aria-label="Close modal"
+					>
+						<?php echo $manifest['resources']['icon']; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>
+					</button>
+				</div>
+			<?php } ?>
+
+			<div class="<?php echo \esc_attr("{$componentClass}__content"); ?>">
```
mbmjertan, line 67: Use the `selector` helper please.

mbmjertan (changes requested):
Just a small comment to use the selector helper for classes in the PHP implementation. Otherwise, looks good!

blocks/init/src/Blocks/components/modal/modal.php
iruzevic: there is no need. If you have one sector there is no need. it is ok to put it like this.

blocks/init/src/Blocks/components/modal/modal.php
iruzevic: there is no need. If you have one sector there is no need. it is ok to put it like this.

blocks/init/src/Blocks/components/modal/modal.php
iruzevic: there is no need. If you have one sector there is no need. it is ok to put it like this.

blocks/init/src/Blocks/components/modal/modal.php
iruzevic: there is no need. If you have one sector there is no need. it is ok to put it like this.

blocks/init/src/Blocks/components/modal/modal.php
iruzevic: there is no need. If you have one sector there is no need. it is ok to put it like this.

mbmjertan (approved):

mbmjertan (approved):

IvanGrginovInf (approved):

PR #480, opened 2021-11-04
Add helper for locking post saving

mbmjertan:
Adds three helpers for locking post saving:
* `lockPostEditing`, locking a post save with a lock handle determined from a block client ID and attribute key
* `unlockPostEditing`, removing a lock created with `lockPostEditing` by removing the lock with the lock handle matching a client ID and attribute key
* `lockIfUndefined`, locking and unlocking a post save automatically if the attribute value is undefined, given a block client ID, attribute key and value

Usage example:
```
			<RichText
				className={headingClass}
				placeholder={placeholder}
				value={headingContent}
				onChange={(value) => { setAttributes({ [getAttrKey('headingContent', attributes, manifest)]: value }); lockIfUndefined(clientId, getAttrKey('headingContent', attributes, manifest), value); }}
				allowedFormats={[]}
				data-id={unique}
			/>
```

I'll do a PR to eightshift-docs documenting this helper as well.


scripts/helpers/lock-editing.js
```diff
@@ -0,0 +1,46 @@
...
+ * A lock is created under the name undefined-lock-[blockClientId]-[attributeKey],
+ * meaning that a lock can be created for each attribute key of each block.
+ * Multiple locks can be present and all of them have to be removed before saving is unlocked.
+ *
+ * @param {string} blockClientId A block's client ID
+ * @param {string} attributeKey The attribute key
+ */
+
+export const lockIfUndefined = (blockClientId, attributeKey, value) => {
+  if (typeof value === 'undefined' || value === null || value === '') {
```
goranalkovic-infinum, line 41: Maybe we shouldn't equate `undefined` and `null` here

dadadavorin (dismissed):
👍  Just solve the linter issues 😄

mbmjertan:
> 👍 Just solve the linter issues 😄

Seems like `npm install` is crashing. Will look into it.

scripts/helpers/index.js
```diff
@@ -19,3 +19,8 @@ export {
 	truncateMiddle,
 	unescapeHTML
 } from './text-helpers';
+export {
+	lockPostEditing,
+	unlockPostEditing,
+	lockIfUndefined
+} from './lock-editing';
```
iruzevic, line 26: this is specific to the editor only not the frontend so please put this in the editor folder not helpers folder and also export order should be alphabetical order. tnx

scripts/helpers/lock-editing.js
```diff
@@ -0,0 +1,46 @@
+/**
```
iruzevic, line 1: please with for all functions

return type, usage and output

like in other helpers
https://github.com/infinum/eightshift-frontend-libs/blob/develop/scripts/editor/props.js

iruzevic (changes requested):
a few small corrections but great job

scripts/helpers/lock-editing.js
mbmjertan: The idea was to lock post saving if there's an attribute that's missing, and I'd wager `null` and an empty string fall under that. I'm not sure what you're suggesting, but I'd suggest using the `lockPostEditing` helper for different use cases.

goranalkovic-infinum (dismissed):

scripts/editor/index.js
```diff
@@ -13,6 +13,11 @@ export {
 } from './get-options';
 export { getPaletteColors } from './get-palette-colors';
 export { inserter } from './inserter';
+export {
+	lockIfUndefined,
+	lockPostEditing,
+	unlockPostEditing
```
volfkarlo, line 19:
```suggestion
	unlockPostEditing,
```

scripts/editor/lock-editing.js
```diff
@@ -0,0 +1,70 @@
...
+ * ```
+ */
+
+export const lockIfUndefined = (blockClientId, attributeKey, value) => {
+  if (typeof value === 'undefined' || value === null || value === '') {
+    lockPostEditing(blockClientId, attributeKey);
+    return;
+  }
+  unlockPostEditing(blockClientId, attributeKey);
+};
```
volfkarlo, line 70: New row at the end.
```suggestion
};

```

volfkarlo (approved):

dingo-d (dismissed):
Looks good to me, just solve Karlo's comments and that's it 😄

goranalkovic-infinum (approved):

dingo-d (approved):

dadadavorin (approved):

iruzevic (approved):

Issue #445, opened 2021-10-04
[FEATURE] ComponentUseToggle Collapsable


iruzevic:
It would be cool if we could make ComponentUseToggle also have collapsible feature like PanelBody component so you can say what it collapsed by default and what is open. I think this will upgrade the UX significantly.

PR #481, opened 2021-11-04
Collapsable ComponentUseToggle

### LINKED ISSUE #461: [FEATURE] Make component options more contained an more user friendly to edit if there are may of them
Some thing like border on hover or something to have better understanding what is what

goranalkovic-infinum:
This is a variation of the `ComponentUseToggle` that we already have (although it's not linked to it, just in case we decide to throw that one out sometimes in the future).

Works mostly like the regular toggle, but also allows collapsing of the options to declutter complex components. There's also an option to have it expanded by default.

Please note that you'll need to put all the component options inside this component (which does declutter and flatten things more!)

Closes #461 and #445.

Example:
Before
```jsx
export const ParagraphEditor = (attributes) => {
	return (
		<>
			<ComponentUseToggle ... />

			// other block options
			// (there are usually many)
		</>
	);
}
```

Now
```jsx
export const ParagraphEditor = (attributes) => {
	return (
		<CollapsableComponentUseToggle ... >
			// other block options
			// (there are usually many)
		</CollapsableComponentUseToggle>
	);
}
```

Here's a video demo of the component in action:
https://user-images.githubusercontent.com/77000136/140386010-e7a58365-05ca-42e9-b1d0-f677313f5f22.mov

### DIFF
--- a/__mocks__/@eightshift/frontend-libs/mocks/wrapper.js
+++ b/__mocks__/@eightshift/frontend-libs/mocks/wrapper.js
@@ -2,4 +2,4 @@ export const Wrapper = (props) => {
 	return (
 		<div>{props.children}</div>
 	);
-}
+};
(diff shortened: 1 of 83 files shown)


iruzevic (dismissed):
❤️

simun-ivanac (approved):

mbmjertan (approved):

PR #484, opened 2021-11-05
Add icon from the parent block to block variations if no icon is supplied

moonbyt3:
# Description

Added check if variation block has icon in it's manifest.
If it doesn't, script will import it's parent manifest and then it will set variation missing icon to parent block icon.

### DIFF
--- a/scripts/editor/register-blocks.js
+++ b/scripts/editor/register-blocks.js
@@ -483,7 +483,14 @@ const registerVariation = (
 ) => {
 
 	// Append globalManifest data in to output.
-	blockManifest['icon'] = getIconOptions(globalManifest, blockManifest);
+	if (blockManifest['icon']) {
+		blockManifest['icon'] = getIconOptions(globalManifest, blockManifest);
+	} else {
+		// There is no icon passed to variation, use it's parent icon instead
+		import(`./../../blocks/init/src/Blocks/custom/${blockManifest.parentName}/manifest.json`).then((parentManifest) => {
+			blockManifest['icon'] = getIconOptions(globalManifest, parentManifest);
+		});
+	}
 
 	// This is a full block name used in Block Editor.
 	const fullBlockName = getFullBlockNameVariation(globalManifest, blockManifest);
--- a/scripts/editor/register-blocks.js
+++ b/scripts/editor/register-blocks.js
@@ -481,7 +481,7 @@ export const getExample = (
 const registerVariation = (
 	globalManifest = {},
 	blockManifest = {},
-	allBlocksManifests
+	allBlocksManifests,
 ) => {
 
 	// Append globalManifest data in to output.
--- a/scripts/editor/register-blocks.js
+++ b/scripts/editor/register-blocks.js
@@ -481,7 +481,7 @@ export const getExample = (
 const registerVariation = (
 	globalManifest = {},
 	blockManifest = {},
-	allBlocksManifests,
+	allBlocksManifests = [],
 ) => {
 
 	// Append globalManifest data in to output.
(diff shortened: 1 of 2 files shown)


scripts/editor/register-blocks.js
```diff
@@ -483,7 +483,14 @@ const registerVariation = (
 ) => {
 
 	// Append globalManifest data in to output.
-	blockManifest['icon'] = getIconOptions(globalManifest, blockManifest);
+	if (blockManifest['icon']) {
+		blockManifest['icon'] = getIconOptions(globalManifest, blockManifest);
+	} else {
+		// There is no icon passed to variation, use it's parent icon instead
```
goranalkovic-infinum, line 489:
```suggestion
		// If no icon is passed to the block variation, use the parent block icon instead.
```

scripts/editor/register-blocks.js
```diff
@@ -483,7 +483,14 @@ const registerVariation = (
...
+		import(`./../../blocks/init/src/Blocks/custom/${blockManifest.parentName}/manifest.json`).then((parentManifest) => {
+			blockManifest['icon'] = getIconOptions(globalManifest, parentManifest);
+		});
```
goranalkovic-infinum, line 492: Consider making the whole method async and then `await`ing the `getIconOptions`.

```suggestion
		const parentManifest = await import(`./../../blocks/init/src/Blocks/custom/${blockManifest.parentName}/manifest.json`);
		blockManifest['icon'] = getIconOptions(globalManifest, parentManifest);
```

scripts/editor/register-blocks.js
```diff
@@ -483,7 +483,14 @@ const registerVariation = (
 ) => {
 
 	// Append globalManifest data in to output.
-	blockManifest['icon'] = getIconOptions(globalManifest, blockManifest);
+	if (blockManifest['icon']) {
+		blockManifest['icon'] = getIconOptions(globalManifest, blockManifest);
+	} else {
+		// There is no icon passed to variation, use it's parent icon instead
+		import(`./../../blocks/init/src/Blocks/custom/${blockManifest.parentName}/manifest.json`).then((parentManifest) => {
```
iruzevic, line 490: this is not the way to go because this path can change in the project and this setup needs to be as generic as possible.

You should pass additional key in the registerVariation function where you will get all blocks, and in that case you just find the icon from the paren key name:

https://github.com/infinum/eightshift-frontend-libs/blob/develop/blocks/init/src/Blocks/assets/scripts/application-blocks-editor.js

iruzevic (changes requested):

goranalkovic-infinum (commented):
Left couple of minor comments, nice job!

scripts/editor/register-blocks.js
iruzevic: the PR approach is wrong so I will remove this comment because it needs to be refactored.

scripts/editor/register-blocks.js
iruzevic: the PR approach is wrong so I will remove this comment because it needs to be refactored.

volfkarlo (changes requested):
👏 

Just a tiny correction/suggestion.

dingo-d (approved):
Looks good to me, just fix the suggestions Karlo left 🙂

scripts/editor/register-blocks.js
```diff
@@ -472,18 +472,29 @@ export const getExample = (
...
+ * @param {object} globalManifest     - Global manifest.
+ * @param {object} blockManifest      - Block manifest.
+ * @param {Array} allBlocksManifests  - All Blocks manifests.
  *
  * @returns {object}
  */
 const registerVariation = (
 	globalManifest = {},
 	blockManifest = {},
+	allBlocksManifests,
```
iruzevic, line 484: please provide the default value so it wound break implementations on the old projects

iruzevic (changes requested):

iruzevic (approved):
looks good

mbmjertan (approved):

dadadavorin (approved):

PR #500, opened 2021-11-22
CustomSelect improvements

goranalkovic-infinum:
Main improvement: **`simpleValue` now also works with async versions of CustomSelect**!

Besides that, new passthrough-able attributes were added:
- `disabled`: if set to `true`, renders the component as disabled.
- `loading`: if set to `true`, renders the component in a loading state.
- `blurInputOnSelect`: if set to `true`, focus is removed from the input once an option is selected.
- `hideSelected`: if set to `true`, the selected option is hidden from the menu.
- `loadingMessage`: text to display when loading options.
- `noOptionsMessage`: text to display when no options are available

### DIFF
--- a/scripts/components/custom-select/custom-select.js
+++ b/scripts/components/custom-select/custom-select.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import Select, { components } from 'react-select';
+import { __ } from '@wordpress/i18n';
 import { useState } from '@wordpress/element';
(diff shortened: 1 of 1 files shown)


dadadavorin (approved):
Nice 👍

simun-ivanac (approved):

iruzevic (approved):

PR #501, opened 2021-11-22
Fix Yoast analysis

### LINKED ISSUE #485: [BUG] Content is saved with unicode escaped strings
When we have a link inside a paragraph component (block) the contents, when code view is toggled, looks something like this

```
<!-- wp:eightshift-boilerplate/paragraph {"wrapperUseSimple":true,"paragraphParagraphContent":"\u003ca href=\u0022https://www.example.com/bla\u0022\u003eBla link\u003c/a\u003eis a \u003cstrong\u003ethis is bolded text\u003c/strong\u003e...
```

Notice the `\u0022` or `\u003c` characters.

Not 100%  sure why this happens, but this is not good, as it will break Yoast analysis tool even with the YoastSEO helper added.

We should see if this can be prevented. That is, the content should be saved as regular UTF-8 strings (< instead of \u003c).

dingo-d:
Improves the regex for capturing content from the attributes. 

I added a way to include image URLs so those are also included in the analysis (when `seo` is added in the image component). 

We should see how to add image `alt` since Yoast probably analyses this as well. I also replaced the unicode characters with their non-unicode counterparts, since that prevented Yoast from detecting outbound and internal links. This was tested on the Yoast Pro version. 

The last thing that probably should get improved is ordering the content because I think this just parses blocks in some weird order. And headings are not wrapped in correct h tags which also skews the analysis.

Partially fixes #485 

This needs to be fixed:

image

This is now fixed:

image

Before it was like this: 

image


iruzevic (dismissed):

dingo-d:
Still need to write tests for this. I'd like to transform our blocks to proper HTML, in order to have a better analysis.

dadadavorin (dismissed):

iruzevic (dismissed):

scripts/plugins/yoast-seo.js
```diff
@@ -11,10 +11,10 @@ class YoastSEOCustomData {
 		}
 
 		// Register custom class.
-			YoastSEO.app.registerPlugin( 'YoastSEOCustomData', { status: 'ready' } );
+		YoastSEO.app.registerPlugin( 'YoastSEOCustomData', { status: 'ready' } );
```
gabriel-glo, line 14:
```suggestion
		YoastSEO.app.registerPlugin('YoastSEOCustomData', {status: 'ready'});
```

gabriel-glo (commented):
Small fix on function arguments.

goranalkovic-infinum (approved):
Approve, but fix the indents Gabriel suggested pls

dingo-d:
I still need to add the tests and the proper HTML parser/serializer. Because there isn't a built-in HTML serializer from the core.

scripts/plugins/yoast-seo.js
```diff
@@ -29,95 +29,95 @@ class YoastSEOCustomData {
 			YoastSEO.app.registerModification( 'content', callback, 'YoastSEOCustomData', 10 );
 	}
 
+
```
volfkarlo, line 32:
```suggestion

```

volfkarlo (changes requested):

dingo-d:
@infinum/wordpress-team anybody willing to take this over? The parser just needs to be added that will chew the content and return HTML (there are automated tests that will help).

iruzevic:
I will make this work

iruzevic:
closing the PR due to different approach.

PR #507, opened 2021-11-25
Lists component improvements

mbmjertan:
This PR:
* implements a way to change the value of the `listsOrdered` attribute from the editor
* implements additional checks that the `listsOrdered` attribute can be only `ul` or `ol`
  * alternatively, we could make the `listsOrdered` attribute a boolean, but I decided for this approach to not break backwards compatibility - let me know if you think we should have a different approach

### DIFF
--- a/blocks/init/src/Blocks/components/lists/components/lists-editor.js
+++ b/blocks/init/src/Blocks/components/lists/components/lists-editor.js
@@ -2,7 +2,7 @@ import React, { useMemo } from 'react';
 import classnames from 'classnames';
 import { __ } from '@wordpress/i18n';
 import { RichText } from '@wordpress/block-editor';
-import { outputCssVariables, getUnique, selector, checkAttr, getAttrKey } from '@eightshift/frontend-libs/scripts';
+import { outputCssVariables, getUnique, selector, checkAttr, getAttrKey, getOption } from '@eightshift/frontend-libs/scripts';
 import manifest from './../manifest.json';
 import globalManifest from './../../../manifest.json';
 
@@ -24,6 +24,12 @@ export const ListsEditor = (attributes) => {
 	const listsUse = checkAttr('listsUse', attributes, manifest);
 	const listsContent = checkAttr('listsContent', attributes, manifest);
 	const listsOrdered = checkAttr('listsOrdered', attributes, manifest);
+
+	const listsOrderedOptions = getOption('listsOrdered', attributes, manifest).map(option => option.value);
+	const listsDefault = manifest.attributes.listsOrdered.default;
+	if (!listsOrderedOptions.includes(listsOrdered)) {
+		setAttributes({ [getAttrKey('listsOrdered', attributes, manifest)]: listsDefault }); // Resets the attribute to the default value if invalid value set.
+	}
 
 	const listsClass = classnames([
 		selector(componentClass, componentClass),
--- a/blocks/init/src/Blocks/components/lists/components/lists-options.js
+++ b/blocks/init/src/Blocks/components/lists/components/lists-options.js
@@ -18,6 +18,7 @@ export const ListsOptions = (attributes) => {
 		showListsColor = true,
 		showListsSize = true,
 		showListsColorOnlyMarker = true,
+		showListsOrdered = true,
 	} = attributes;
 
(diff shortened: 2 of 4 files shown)


blocks/init/src/Blocks/components/lists/components/lists-options.js
```diff
@@ -62,6 +64,18 @@ export const ListsOptions = (attributes) => {
 						/>
 					}
 
+					{showListsOrdered &&
+						<CustomSelect
```
goranalkovic-infinum, line 68: As this is less than 5 options, a `SimpleVerticalSingleSelect` would be a better option here.

There are icons for (un)ordered lists you can use 😉 
image

blocks/init/src/Blocks/components/lists/manifest.json
```diff
@@ -38,6 +38,16 @@
 		}
 	},
 	"options": {
+		"listsOrdered": [
+			{
+				"label": "Ordered list",
+				"value": "ol"
+			},
+			{
+				"label": "Bullet list",
```
goranalkovic-infinum, line 47:
```suggestion
				"label": "Bulleted list",
```

blocks/init/src/Blocks/components/lists/components/lists-editor.js
```diff
@@ -25,6 +25,12 @@ export const ListsEditor = (attributes) => {
 	const listsContent = checkAttr('listsContent', attributes, manifest);
 	const listsOrdered = checkAttr('listsOrdered', attributes, manifest);
 
+	const listsOrderedOptions = getOption('listsOrdered', attributes, manifest).map(option => option.value);
+	const listsDefault = manifest.attributes.listsOrdered.default;
+	if (!listsOrderedOptions.includes(listsOrdered)) {
+		setAttributes({ [getAttrKey('listsOrdered', attributes, manifest)]: listsDefault }); // Resets the attribute to the default value if invalid value set.
```
goranalkovic-infinum, line 31: Is there a way to force an invalid value besides manually editing the code view?

If not, maybe this check should be only on PHP side

goranalkovic-infinum (changes requested):
Left a couple of Qs and suggestions, nice job!

blocks/init/src/Blocks/components/lists/components/lists-editor.js
mbmjertan: There isn't, but `RichText` is rendered with that HTML tag, so I'm not too sure we should render an attribute we're not sure is safe.

blocks/init/src/Blocks/components/lists/components/lists-options.js
```diff
@@ -62,6 +64,18 @@ export const ListsOptions = (attributes) => {
 						/>
 					}
 
+					{showListsOrdered &&
+						<SimpleVerticalSingleSelect
+							label={__('List type', 'edp')}
```
goranalkovic-infinum, line 69:
```suggestion
							label={__('List type', 'eightshift-frontend-libs')}
```

blocks/init/src/Blocks/components/lists/manifest.json
```diff
@@ -38,6 +38,18 @@
 		}
 	},
 	"options": {
+		"listsOrdered": [
+			{
+				"label": "Ordered list",
```
goranalkovic-infinum, line 43:
```suggestion
				"label": "Ordered",
```

blocks/init/src/Blocks/components/lists/manifest.json
```diff
@@ -38,6 +38,18 @@
...
+				"value": "ol",
+				"icon": "listOrdered"
+			},
+			{
+				"label": "Bulleted list",
```
goranalkovic-infinum, line 48:
```suggestion
				"label": "Bulleted",
```

goranalkovic-infinum (changes requested):
Couple more for cohesion 😄

dingo-d (dismissed):
Just fix what Goc suggested and it's ok from my end 👍🏼

goranalkovic-infinum (approved):

simun-ivanac (approved):

PR #530, opened 2021-12-14
adding overrides option in block registration

iruzevic:
# Description

This file gives you the option to any override block manifest value by providing the same key.
With this feature you can pass anything from PHP using translation variables to JS and do some kind of conditions with that.

# Screenshots / Videos
Screenshot 2021-12-14 at 11 59 05


goranalkovic-infinum (approved):

volfkarlo (approved):
💪

PR #535, opened 2022-01-11
Classnames helper

gabriel-glo:
# Description

Replaced the `classnames` helper function from classnames npm package with custom helper that's more similar to the one from [eightshift-libs](https://github.com/infinum/eightshift-libs/blob/develop/src/Helpers/Components.php#L60).

As this is a breaking change, we'll probably need to postpone this to a later version.


scripts/helpers/classnames.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Converts an array of classes into a string.
+ *
+ * @param {array} classnamesArray Array of classes.
+ *
+ * @return string
+ */
+export function classnames(classnamesArray) {
+	return classnamesArray.filter(String).join(' ').trim();
```
goranalkovic-infinum, line 9: Maybe we could also trim each class.

```suggestion
	return classnamesArray.filter(String).map((class) => class.trim()).join(' ').trim();
```

and also remove empty strings.
```suggestion
	return classnamesArray.filter(String).map((class) => class.trim()).filter((class) => class.length).join(' ').trim();
```

goranalkovic-infinum (changes requested):
Left 1 comment.

I wouldn't call this a breaking change if (for now) the old npm package stays in the project.

dingo-d (changes requested):
Can you write some unit tests for this helper?

Try to cover the cases when you need to add single string, multiple strings, arrays, conditional inside the classnames array, etc.

scripts/helpers/classnames.js
gabriel-glo: `.filter(String)` already removes empty strings in the array. 
Maybe we can first trim each string in the array, then remove empty strings and then join all classes into a single string?

package.json
```diff
@@ -56,7 +56,6 @@
 		"babel-loader": "^8.2.2",
 		"browser-sync": "^2.26.14",
 		"browser-sync-webpack-plugin": "^2.3.0",
-		"classnames": "^2.2.6",
```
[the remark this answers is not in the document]
goranalkovic-infinum: That should be fine tho, if you have a project that still uses it, it won't be treeshaken-out. If you're in a "new" project, it won't be there.

tests/unit/scripts/helpers/classnames.test.js
```diff
@@ -0,0 +1,43 @@
...
+		'',
+		' ',
+		selector(!accordionManifest.componentClass, accordionManifest.componentClass),
+		selector(!accordionManifest.componentJsClass, accordionManifest.componentJsClass),
+	]);
+
+	expect(accordionClassnames).toBe('');
+});
+
+it('tests classnames returns only valid classes', () => {
```
dingo-d, line 29: Can you add a test where you have a ternary as one of the arguments in the classnames array? Just so that we cover all the cases. 

Also, can you think of any 'unhappy path' tests - the test where you give it invalid input and expect a correct error?

Thanks!

mbmjertan: Just gonna leave this here :eyes: https://stackoverflow.com/questions/448981/which-characters-are-valid-in-css-class-names-selectors

gabriel-glo: I've added additional case when ternary operator is used as an item in class names array.

As for the invalid CSS class names, I think this helper shouldn't worry about that, since this is not its domain.
The purpose of this helper is to create a string of class names out of array of potential class names. Invalid classnames have more to do with CSS for styling and JS for adding interactions.  

Also how would this be tested? Running a `document.querySelector()` method and evaluating the results seems kind of redundant and a large overhead in this case.

dingo-d (dismissed):

mbmjertan (approved):

iobrado (approved):

iruzevic:
Could you please change this to release/5.0.0 and fix merge conflicts? tnx

mbmjertan (approved):

gabriel-glo:
@infinum/wordpress-team 
What's going on with this conflict in this docblock? There is no way to resolve it in my editor and git tells me that it's already synced.

dingo-d (approved):

goranalkovic-infinum (approved):

blocks/init/src/Blocks/components/lists/components/lists-editor.js
```diff
@@ -1,8 +1,8 @@
 import React, { useMemo } from 'react';
-import classnames from 'classnames';
 import { __ } from '@wordpress/i18n';
 import { RichText } from '@wordpress/block-editor';
-import { outputCssVariables, getUnique, selector, checkAttr, getAttrKey, getOption } from '@eightshift/frontend-libs/scripts';
+import { outputCssVariables, getUnique, selector, checkAttr, getAttrKey, getOption, classnames
+ } from '@eightshift/frontend-libs/scripts';
```
volfkarlo, line 5: looks a bit strange maybe to put each in a separate row?

```suggestion
import {
outputCssVariables,
getUnique,
selector,
checkAttr,
getAttrKey,
getOption,
classnames,
} from '@eightshift/frontend-libs/scripts';
```

volfkarlo (changes requested):
Minor change.

Also, I've noticed that you haven't added the script to scritps/index.js which means that you can't jump directly to the function.

blocks/init/src/Blocks/components/lists/components/lists-editor.js
gabriel-glo: @volfkarlo  But you did add the helper in `scripts/index.js`, so there is no need for me to do anything about it?

volfkarlo: Did I? 😂 

If it works awesome 😄

PR #540, opened 2022-01-18
Improve header / footer accessibility, fix #534

### LINKED ISSUE #534: Drawer A11y issues
### Describe your bug

Currently, eightshift-frontend-libs provides a drawer component which uses a `transform: translate3d(100%, 0, 0);` rule (with `display: block;`) to hide itself and animate opening/closing. It also uses a page overlay component which renders a `.page-overlay` div to darken the rest of the page (and allow the drawer to be closed by clicking on the page itself)

However, this approach introduces some A11y issues:

- [ ] As the drawer has `display: block`, its elements are part of the tab order, even in cases where a drawer trigger isn't rendered (such as on desktop) or drawer is closed - this creates confusion for keyboard users
- [ ] Clickable elements must be focusable and should have interactive semantics - https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Keyboard

Generally speaking, keyboard control could be improved here (e.g., supporting `esc` to close the drawer)

### Steps to Reproduce

1. Set up a menu on the default eightshift-boilerplate.

### Expected behavior

Drawer elements aren't part of the tab order when the drawer is closed. Drawer is fully keyboard accessible. 

### Screenshots, screen recording, code snippet

Screenshot 2022-01-10 at 10 14 42
Screenshot 2022-01-10 at 10 15 14

### Environment info

_No response_

### Please confirm that you have searched existing issues in this repo.

Yes

### Please confirm that you have searched in our documentation and couldn't find the answer.

Yes

### Please confirm that your bug occurs with all plugins uninstalled and with the default WordPress theme active.

No

mbmjertan:
# Description

This PR, [along with the one on infinum/eightshift-boilerplate](https://github.com/infinum/eightshift-frontend-libs/pull/540), fixes the accessibility issues reported in #534.

* `layout-three-columns` now supports setting an ARIA role for it
* `hamburger` now supports labels
* `drawer` has been refactored to improve accessibility and make the component simpler:
  * page overlay and positioning support was removed, which fixes the interactive semantics issues with page overlays, and also IMO looks a bit cleaner
  * drawers now use `display: none` to hide themselves, instead of visibility hacks, which fixes the tab order issues

# Screenshots / Videos

The new drawer:
Screenshot 2022-01-18 at 10 10 29

Tab order demo:

https://user-images.githubusercontent.com/1742806/149909699-38b292c6-b9fa-4da6-9c8c-7e475b85ef51.mov

### DIFF
--- a/blocks/init/src/Blocks/components/layout-three-columns/layout-three-columns.php
+++ b/blocks/init/src/Blocks/components/layout-three-columns/layout-three-columns.php
@@ -24,6 +24,7 @@ $layoutThreeColumnsLeft = Components::checkAttr('layoutThreeColumnsLeft', $attri
 $layoutThreeColumnsCenter = Components::checkAttr('layoutThreeColumnsCenter', $attributes, $manifest);
 $layoutThreeColumnsRight = Components::checkAttr('layoutThreeColumnsRight', $attributes, $manifest);
 $layoutThreeColumnsHtmlTag = Components::checkAttr('layoutThreeColumnsHtmlTag', $attributes, $manifest);
+$layoutThreeColumnsAriaRole = Components::checkAttr('layoutThreeColumnsAriaRole', $attributes, $manifest);
 
 $layoutClass = Components::classnames([
 	Components::selector($componentClass, $componentClass),
@@ -51,7 +52,12 @@ $columnRightClass = Components::classnames([
 
 ?>
 
-<<?php echo esc_attr($layoutThreeColumnsHtmlTag); ?> class="<?php echo \esc_attr($layoutClass); ?>">
+<<?php echo esc_attr($layoutThreeColumnsHtmlTag); ?>
+	class="<?php echo \esc_attr($layoutClass); ?>"
+	<?php if ($layoutThreeColumnsAriaRole) { ?>
+		role="<?php echo \esc_attr($layoutThreeColumnsAriaRole); ?>"
+	<?php } ?>
+>
 
 		<?php if ($layoutThreeColumnsLeft) { ?>
 			<div class="<?php echo \esc_attr($columnLeftClass); ?>">
--- a/blocks/init/src/Blocks/components/layout-three-columns/manifest.json
+++ b/blocks/init/src/Blocks/components/layout-three-columns/manifest.json
@@ -24,6 +24,10 @@
 			"type": "string",
 			"default": "div"
 		},
+		"layoutThreeColumnsAriaRole": {
+			"type": "string",
+			"default": ""
+		},
 		"layoutThreeColumnsUse": {
 			"type": "boolean",
 			"default": true
--- a/blocks/init/src/Blocks/components/layout-three-columns/manifest.json
+++ b/blocks/init/src/Blocks/components/layout-three-columns/manifest.json
@@ -25,8 +25,7 @@
 			"default": "div"
 		},
 		"layoutThreeColumnsAriaRole": {
-			"type": "string",
-			"default": ""
+			"type": "string"
 		},
 		"layoutThreeColumnsUse": {
 			"type": "boolean",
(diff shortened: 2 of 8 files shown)


dadadavorin (commented):
Code seems fine - but what about animations, when you are hovering the hamburger and when opening the mobile menu? These are pretty nice and useful basic functionality (and on some projects they are used the way they are).

mbmjertan:
@dadadavorin: 

With this PR, the hamburger button itself is animated, while the opening of the drawer isn't - as you can notice in the video. Generally, the hamburger wasn't changed.

We could still animate this like so:
* use `display: none;` to hide the drawer element
* position it outside the viewport (using transforms or position)
* set the drawer to `display: block` on click and add a debounced function to position it in the viewport, animating that

Didn't understand this was a requirement for the drawer, and I wanted to make it as simple as possible. If the rest of the team agrees that this would be beneficial, I can animate the drawer as well.

dingo-d (dismissed):

blocks/init/src/Blocks/components/layout-three-columns/layout-three-columns.php
```diff
@@ -51,7 +52,12 @@
 
 ?>
 
-<<?php echo esc_attr($layoutThreeColumnsHtmlTag); ?> class="<?php echo \esc_attr($layoutClass); ?>">
+<<?php echo esc_attr($layoutThreeColumnsHtmlTag); ?>
+	class="<?php echo \esc_attr($layoutClass); ?>"
+	<?php if ($layoutThreeColumnsAriaRole) { ?>
+		role="<?php echo \esc_attr($layoutThreeColumnsAriaRole); ?>"
```
gabriel-glo, line 58: Is `role` attribute something that can be left with an empty or some default value?
I'm thinking that this conditional can be written with a ternary or null coalesce operators.

It will look a bit cleaner IMO.

blocks/init/src/Blocks/components/layout-three-columns/manifest.json
```diff
@@ -24,6 +24,10 @@
 			"type": "string",
 			"default": "div"
 		},
+		"layoutThreeColumnsAriaRole": {
+			"type": "string",
+			"default": ""
```
gabriel-glo, line 29: Is `default` property needed if it's empty?

dadadavorin (dismissed):

gabriel-glo (changes requested):
Left a few questions.

I think you shouldn't just remove the animation of the drawer. There is no serious project where that will be fine. Try to have just one animation in right positioned drawer. You'd need to defer "animation" of display property on the drawer.

blocks/init/src/Blocks/components/layout-three-columns/layout-three-columns.php
mbmjertan: It should fall back to the implicit role if empty. However, the ARIA specification advises against adding redundant `role` attributes, which is why I used this. 

> The explicit semantic role of an element is determined by its role attribute (if any).
> 
> The role attribute takes a list of tokens. The explicit semantic role is the first valid role in this list. The valid roles are all non-abstract roles from WAI-ARIA Specifications. If the element has no role attribute, or if it has one with no valid role, then this element has no explicit semantic role.
> 
> The implicit semantic role of an element is a pre-defined value given by the host language which depends on the element and its ancestors.
> 
> (from https://www.w3.org/WAI/standards-guidelines/act/rules/role-attribute-valid-value-674b10/)

blocks/init/src/Blocks/components/layout-three-columns/manifest.json
iruzevic: @mbmjertan can you fix this so we can merge this pr?

mbmjertan: Will do, @iruzevic! :)

iruzevic (approved):

dingo-d (approved):

iobrado (approved):

dadadavorin (approved):

PR #558, opened 2022-02-08
Accordion A11y improvements

mbmjertan:
# Description

Fixes some a11y issues with the accordion component:

* setting ARIA attributes using `Element` properties isn't supported in Firefox, so this was changed to `setAttribute`
* the accordion panel region didn't have the `aria-labelledby` attribute
* when clicking on the accordion, text would flash between black and white a bit because the hover state changed it to white, but the active state had `color: inherit;` set - which would inherit the color disregarding the hover state

# Screenshots / Videos

# Linked documentation PR

### DIFF
--- a/blocks/init/src/Blocks/components/accordion/assets/accordion.js
+++ b/blocks/init/src/Blocks/components/accordion/assets/accordion.js
@@ -63,8 +63,8 @@ export class Accordion {
 		}
 
 		item.dataset.accordionOpen = false;
-		item.querySelector(this.triggerSelector)?.setAttribute("aria-expanded", "false");
-		panel.setAttribute("aria-hidden", "true");
+		item.querySelector(this.triggerSelector)?.setAttribute('aria-expanded', 'false');
+		panel.setAttribute('aria-hidden', 'true');
 	}
 
 	open(item, panel) {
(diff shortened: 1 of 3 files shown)


iobrado (dismissed):

blocks/init/src/Blocks/components/accordion/assets/accordion.js
```diff
@@ -63,8 +63,8 @@ export class Accordion {
 		}
 
 		item.dataset.accordionOpen = false;
-		item.ariaExpanded = false;
-		panel.ariaHidden = true;
+		item.querySelector(this.triggerSelector)?.setAttribute("aria-expanded", "false");
+		panel.setAttribute("aria-hidden", "true");
```
dingo-d, line 67: Why double quotation instead of single?

dingo-d (changes requested):
Just left one question

blocks/init/src/Blocks/components/accordion/assets/accordion.js
mbmjertan: No particular reason, just a typo.

iobrado (approved):

goranalkovic-infinum (approved):

dingo-d (approved):

PR #559, opened 2022-02-08
Change color scheme to match WCAG contrast requirements

mbmjertan:
# Description

This PR changes the default color scheme so the contrast with white or black foregrounds is in line with WCAG AAA requirements (as they're mostly used as background colors for white or black text, mostly on buttons and as wrapper backgrounds).

As this is WCAG AAA compliant, lesser contrast (which will probably happen if the foreground colors aren't exactly `#fff` or `#000`) should still be AA-level compliant.

As project usually define their own color schemes, this PR will only affect the starter theme. As discussed previously with @iruzevic, these colors don't have brand meaning and were chosen arbitrarily, so these changes are mostly the same.

I'll work on a contrast warning later on.

# Screenshots / Videos
This is what buttons now look like:
Screenshot 2022-02-08 at 15 28 56

# Linked documentation PR

Relevant: https://webaim.org/resources/contrastchecker/

### DIFF
--- a/blocks/init/src/Blocks/manifest.json
+++ b/blocks/init/src/Blocks/manifest.json
@@ -40,7 +40,7 @@
 			{
 				"name": "Primary",
 				"slug": "primary",
-				"color": "#662ED6"
+				"color": "#6200EA"
 			},
 			{
 				"name": "Black",
@@ -60,42 +60,42 @@
 			{
 				"name": "Ocean",
 				"slug": "ocean",
-				"color": "#155D7E"
+				"color": "#00B8D4"
 			},
 			{
 				"name": "Green",
 				"slug": "green",
-				"color": "#175959"
+				"color": "#00E676"
 			},
 			{
 				"name": "Red",
 				"slug": "red",
-				"color": "#AF1818"
+				"color": "#B30024"
 			},
 			{
 				"name": "Yellow",
 				"slug": "yellow",
-				"color": "#77460D"
+				"color": "#FFEA00"
 			},
 			{
 				"name": "Blue",
 				"slug": "blue",
-				"color": "#285786"
+				"color": "#0040F0"
 			},
 			{
 				"name": "Mist",
 				"slug": "mist",
-				"color": "#F0F8FF"
+				"color": "#F1FBFE"
 			},
 			{
 				"name": "Dew",
 				"slug": "dew",
-				"color": "#FBF9FF"
+				"color": "#FAF8FC"
 			},
 			{
 				"name": "Cream",
 				"slug": "cream",
-				"color": "#F4F3F0"
+				"color": "#FFFBF5"
 			},
 			{
 				"name": "Eightshift",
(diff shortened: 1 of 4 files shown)


iobrado (dismissed):

blocks/init/src/Blocks/manifest.json
```diff
@@ -60,27 +60,27 @@
...
 				"name": "Red",
 				"slug": "red",
-				"color": "#E84E4E"
+				"color": "#AF1818"
 			},
 			{
 				"name": "Yellow",
 				"slug": "yellow",
-				"color": "#F99922"
+				"color": "#77460D"
```
goranalkovic-infinum, line 78:
```suggestion
				"color": "#FFEA00"
```
Needs to have black text for WCAG AAA.

blocks/init/src/Blocks/manifest.json
```diff
@@ -60,27 +60,27 @@
...
 				"name": "Green",
 				"slug": "green",
-				"color": "#29A3A3"
+				"color": "#175959"
 			},
 			{
 				"name": "Red",
 				"slug": "red",
-				"color": "#E84E4E"
+				"color": "#AF1818"
```
goranalkovic-infinum, line 73:
```suggestion
				"color": "#B30024"
```

blocks/init/src/Blocks/manifest.json
```diff
@@ -60,27 +60,27 @@
 			{
 				"name": "Ocean",
 				"slug": "ocean",
-				"color": "#44AEDE"
+				"color": "#155D7E"
```
goranalkovic-infinum, line 63:
```suggestion
				"color": "#00B8D4"
```
Needs to have black text for WCAG AAA.

gabriel-glo (approved):

goranalkovic-infinum (changes requested):
While the current colors aren't any official brand colors, they represented all the color of Eightshift and its projects 😄 

But to be fully a11y compliant, here are some new colors that are fully WCAG AAA compliant. (Some of the current colors like yellow were compared to a white foreground, there was never a chance they would be WCAG AAA compliant 😄 , the ones that need a black foreground to be compliant are highlighted in change requests).

Also, here are new values for:
- `mist`: `#F1FBFE`
- `dew`: `#FAF8FC`
- `cream`: `#FFFBF5`

Here's a preview:
image

blocks/init/src/Blocks/manifest.json
```diff
@@ -40,7 +40,7 @@
 			{
 				"name": "Primary",
 				"slug": "primary",
-				"color": "#9973E3"
+				"color": "#662ED6"
```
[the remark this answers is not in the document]
mbmjertan: Isn't this #6F2ADE on the screenshot?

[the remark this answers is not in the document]
goranalkovic-infinum: Good catch, should be #662ED6

mbmjertan:
Changes suggested by @goranalkovic-infinum have been added, and I've went ahead and set some different defaults for link colors and their hover states in paragraphs and lists for contrast improvements.

goranalkovic-infinum (approved):

dingo-d (approved):

PR #562, opened 2022-02-15
Add skip link component

mbmjertan:
# Description

This PR adds a `skip-link` component, which provides a hidden, yet focusable skip link.

In its default configuration, it provides a skip link to `#main-content`. (https://github.com/infinum/eightshift-boilerplate/pull/253).

It's not added to the header itself, and it doesn't set a tabindex, as to support using the component to implement [G1](https://www.w3.org/WAI/WCAG21/Techniques/general/G1.html), [G123](https://www.w3.org/WAI/WCAG21/Techniques/general/G123.html) and [G124](https://www.w3.org/WAI/WCAG21/Techniques/general/G124.html) techniques. The technique used is left as a choice for developers. Also, we provide nice landmarks by default, so WCAG doesn't require us to use skip links. 

However, this might differ in the real world, so that's why this component exists.

# Screenshots / Videos

https://user-images.githubusercontent.com/1742806/154076445-84b49f8b-76e7-47d8-9408-7bf15232d1de.mov

# Linked documentation PR

See component readme.


blocks/init/src/Blocks/components/skip-link/skip-link-style.scss
```diff
@@ -0,0 +1,12 @@
...
+	top: -1000px;
+	background-color: var(--global-colors-mist);
+	color: var(--global-colors-black);
+	padding: 0.3em;
+
+	&:focus {
+		top: 10px;
+		left: 10px;
+	}
+}
```
iobrado, line 12: An extra blank line here would be nice 😄

iobrado (approved):

iobrado (approved):

dadadavorin (approved):

PR #564, opened 2022-02-16
Improvements for small viewports

mbmjertan:
# Description

This PR is focused on preventing overflows and clipping in regards to [WCAG AA SC 1. 4. 10.: Reflow](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html).

* Makes the number of items per row in the Featured Posts block a responsive attribute, which was previously a single value for all viewports.
* Stops the header component from overflowing by setting `min-height` instead of `height`
* Makes use of `flex-wrap` in `layout-three-columns` and `menu-horizontal` so it doesn't clip or overflow

# Screenshots / Videos

Wrapping:
Screenshot 2022-02-16 at 10 31 40

Screenshot 2022-02-16 at 10 32 11

Responsive Featured posts:
https://user-images.githubusercontent.com/1742806/154236174-898535ae-6f24-41cf-b3fb-f30d0d14e283.mov

Non-overflowing content:
Screenshot 2022-02-16 at 10 35 52
Screenshot 2022-02-16 at 10 37 01

# Linked documentation PR

### DIFF
--- a/blocks/init/src/Blocks/components/header/header-style.scss
+++ b/blocks/init/src/Blocks/components/header/header-style.scss
@@ -9,7 +9,7 @@
 
 	width: 100%;
 	max-width: var(--global-containers-default);
-	height: var(--header-scoped-height, 5rem);
+	min-height: var(--header-scoped-height, 5rem);
 
 	background: var(--header-scoped-background, var(--global-colors-white));
 
--- a/blocks/init/src/Blocks/components/heading/heading-style.scss
+++ b/blocks/init/src/Blocks/components/heading/heading-style.scss
@@ -4,4 +4,5 @@
 	color: var(--heading-color, var(--global-colors-black, currentColor));
 	font-size: var(--heading-font-size, 1.5rem);
 	line-height: var(--heading-line-height, 1.25);
+	overflow-wrap: anywhere;
 }
--- a/blocks/init/src/Blocks/components/layout-three-columns/layout-three-columns-style.scss
+++ b/blocks/init/src/Blocks/components/layout-three-columns/layout-three-columns-style.scss
@@ -2,4 +2,6 @@
 	display: flex;
 	justify-content: space-between;
 	align-items: center;
+	flex-wrap: wrap;
+	gap: 0.75rem;
 }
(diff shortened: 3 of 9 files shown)


dingo-d (commented):
Should we also add a `word-break: break-word` to heading, paragraph and menu items by default to prevent content overflow?

mbmjertan:
The `break-word` value is deprecated for `word-break`.

The only case where there can actually be an overflow due to content length (when using the default line breaking rules) is when you have `wordswiderthantheviewport`. I generally tend to avoid using non-default line breaking rules as Chrome has a kinda bad algorithm when determining whether to break a word or reflow, but I've played around with this a little and seems that they've improved upon it.

iobrado (dismissed):

gabriel-glo (dismissed):

dingo-d:
> The break-word value is deprecated for word-break.

This is the first I've hear of it. 

https://caniuse.com/mdn-css_properties_word-break_break-word
https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

No mention of deprecations that I can see 👀 

On almost all projects I encounter issues where somebody pastes a super long URL and it goes through the container, sometimes on desktop, more on mobile, so just adding break-word fixed that issue.

mbmjertan:
> No mention of deprecations that I can see 👀

From the MDN link you've left 😄 
Screenshot 2022-02-16 at 14 11 18

> On almost all projects I encounter issues where somebody pastes a super long URL and it goes through the container, sometimes on desktop, more on mobile, so just adding break-word fixed that issue.

A valid point! I'll work on this.

dingo-d:
I'll be damned: https://rocketvalidator.com/html-validation/css-the-value-break-word-is-deprecated#:~:text=The%20value%20break%2Dword%20for,it%20with%20a%20valid%20value.&text=For%20compatibility%20with%20legacy%20content,a%20deprecated%20break%2Dword%20keyword.

Wasn't aware of that. But it looks like it will do automatic fall back 🤷🏼‍♂️

iobrado (approved):

dingo-d (approved):

PR #574, opened 2022-03-02
Fixes for Quote block/component a11y

mbmjertan:
# Description

The `quote` component currently renders the quote content as a `heading`. 

This is problematic as the `heading` component always outputs an `h1-h6` element, and assistive technologies allow users to navigate the page using them. Using headings for quotes (and other non-heading content) poisons our accessibility tree.

To fix this, I've replaced the use of the `heading` component with the `paragraph` component inside the `quote` component. However, to make sure that this component switcheroo isn't a breaking change, I've kept the `heading` prefix for quote content. 

Also, to make sure that quotes don't look too different from what they've looked like before, and so that configuration attributes aren't lost, I've added the options we have for `headingSize` to the `paragraph` component as well. The styles and sizes for them aren't a 1-to-1 match (because existing options already differ), but all options exists so that they don't get overridden if someone updates to this component version. 

To improve the user experience a bit, I've also added some labels and switches to `ParagraphOptions` in `QuoteOptions`.

# Screenshots / Videos

Look, ma, I'm a `<p>`.
Screenshot 2022-03-02 at 15 48 56

Who turned off the attribution?
Screenshot 2022-03-02 at 15 48 24


blocks/init/src/Blocks/components/quote/components/quote-options.js
```diff
@@ -36,13 +36,18 @@ export const QuoteOptions = (attributes) => {
...
+					<ParagraphOptions
+						label={__('Quote', 'eightshift-frontend-libs')}
+						showLabel={true}
 						{...props('heading', attributes, {
 							options: getOptions(attributes, manifest),
 						})}
 					/>
 
 					<ParagraphOptions
+						label={__('Attribution', 'access-guru')}
```
iobrado, line 48:
```suggestion
						label={__('Attribution', 'eightshift-frontend-libs')}
```

iobrado (changes requested):
I've noticed one wrong textdomain 🙂

iobrado (approved):

dingo-d (approved):

dadadavorin (approved):

PR #575, opened 2022-03-03
Global CSS variables color values output

goranalkovic-infinum:
# Description

When we run into a situation where we need a project color with opacity, we had to resort to going to the manifest, getting that color's hex code, then either converting it to RGB (so we can do a transparent RGB) or converting 0-255 to hex and adding it at the end of the hex value.

Because that process is annoying and make maintainability worse (you don't have a link to the original color, so if it ever changes you're in trouble), this PR adds a "color value" output alongside the regular color output we had right now.

Now, if you need a transparent color it's easier than ever!

Let's say you have a color in the manifest named `primary`. If you need set that color for text for example, as usual you would just use:
```css
color: var(--global-colors-primary);
```

If you need that color with 45% opacity, now you can do:
```css
color: rgb(var(--global-colors-primary-values) / 0.45);
```
or
```css
color: rgb(var(--global-colors-primary-values) / 45%);
```
(basically just copy the "regular" variable that was used, add **`-values`** to its name, and wrap it with `rgb()` and add `/ <opacity>` at the end.

**⚠️ Note**
**Only the modern, space-separated `rgb()` syntax is supported (`rgb(<red> <green> <blue> [/ <alpha>])`**, the legacy one with comma-separated values and the `rgba()` syntax (`rgb(<red>, <green>, <blue>)` / `rgba(<red>, <green>, <blue>, <alpha>)`) **will not work**.

# Screenshots / Videos

\-

# Linked documentation PR

\-

### DIFF
--- a/tests/unit/scripts/editor/hex-to-rgb.test.js
+++ b/tests/unit/scripts/editor/hex-to-rgb.test.js
@@ -1,0 +1,12 @@
+import { hexToRgb } from "../../../../scripts/editor";
+
+test.each([
+	{input: '', expected: '0 0 0'},
+	{input: '#fff', expected: '255 255 255'},
+	{input: 'invalid', expected: '0 0 0'},
+	{input: '#ooolll', expected: '0 0 0'},
+	{input: '#', expected: '0 0 0'},
+	{input: '#123456', expected: '18 52 86'},
+])('tests hextToRgb returns a valid value %s', ({input, expected}) => {
+	expect(hexToRgb(input)).toBe(expected);
+});


mbmjertan (dismissed):
Very clean conversion approach 👍

simun-ivanac (dismissed):

iobrado (dismissed):

dadadavorin (approved):
Nice 👍

scripts/editor/output-css-variables.js
```diff
@@ -117,6 +117,32 @@ export const outputCssVariablesGlobal = (globalManifest) => {
...
+		const [r1, g1, b1] = h;
+		r = `0x${r1}${r1}`;
+		g = `0x${g1}${g1}`;
+		b = `0x${b1}${b1}`;
+	} else if (h.length === 6) {
+		const [r1, r2, g1, g2, b1, b2] = h;
+		r = `0x${r1}${r2}`;
+		g = `0x${g1}${g2}`;
+		b = `0x${b1}${b2}`;
+	}
```
dingo-d, line 141: What happens if the `h` length is something different than 3 or 6? I don't see an error here. Also, what does the `h` stands for?

dingo-d (commented):
Left one question.

Also, can you write some unit tests for this? It's a util helper so it shouldn't be hard to do 🙂

scripts/editor/output-css-variables.js
goranalkovic-infinum: `h` = hex code

Tweaked the variable name now so it's more clear 😄

dingo-d (approved):

mbmjertan (approved):

PR #585, opened 2022-03-10
Fix lint errors after Eightshift Coding Standards 1.4 release

mbmjertan:
# Description

I ran `composer standards:fix` and replaced `WordPress.Security.EscapeOutput.OutputNotEscaped` ignores with `Eightshift.Security.ComponentsEscape.OutputNotEscaped`.

### DIFF
--- a/blocks/init/src/Blocks/components/accordion/accordion.php
+++ b/blocks/init/src/Blocks/components/accordion/accordion.php
@@ -53,33 +53,33 @@ $uniqueAccordionTriggerId = Components::getUnique();
 ?>
 
 <div
-	class="<?php echo \esc_attr($accordionClass); ?>"
-	data-accordion-open="<?php echo \esc_attr($accordionIsOpen ? 'true' : 'false'); ?>"
-	data-close-others="<?php echo \esc_attr($accordionCloseOthers ? 'true' : 'false'); ?>"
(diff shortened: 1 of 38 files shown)


blocks/init/src/Blocks/components/accordion/accordion.php
```diff
@@ -53,33 +53,33 @@
 ?>
 
 <div
-	class="<?php echo \esc_attr($accordionClass); ?>"
-	data-accordion-open="<?php echo \esc_attr($accordionIsOpen ? 'true' : 'false'); ?>"
-	data-close-others="<?php echo \esc_attr($accordionCloseOthers ? 'true' : 'false'); ?>"
+	class="<?php echo esc_attr($accordionClass); ?>"
```
dingo-d, line 56: This is odd tbh because the slevomat CS should enforce the fully qualified global functions and constants.

So if you remove the slashes, you should import them at the top with `use` statement.

The rationale is to optimize these calls.

mbmjertan: Screenshot 2022-03-10 at 15 22 04

dingo-d: Oh totally missed that. Neat! Ignore me then XD

dingo-d (approved):

PR #610, opened 2022-04-27
Fix Gradient Picker component

volfkarlo:
# Description
- add GradientPicker component with fallback for versions prior to 5.9

# Screenshots / Videos
gradient working on 5 9

### DIFF
--- a/scripts/components/advanced-color-picker/advanced-color-picker.js
+++ b/scripts/components/advanced-color-picker/advanced-color-picker.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import { __ } from '@wordpress/i18n';
-import { BaseControl, ColorPicker, __experimentalGradientPicker as GradientPicker } from '@wordpress/components';
+import { BaseControl, ColorPicker, GradientPicker, __experimentalGradientPicker as GradientPickerOld } from '@wordpress/components';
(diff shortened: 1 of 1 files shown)


goranalkovic-infinum (approved):

iobrado (approved):

scripts/components/advanced-color-picker/advanced-color-picker.js
```diff
@@ -138,7 +140,7 @@ export const AdvancedColorPicker = (props) => {
 			}
 
 			{type === 'gradient' && showGradient && !disabled &&
-				<GradientPicker
+				<GradientPickerComponent
```
iruzevic, line 143: what are the changes implemented in the newer version so we need to reference 2 different gradient component?
Are the only names changed or?

iruzevic (changes requested):

scripts/components/advanced-color-picker/advanced-color-picker.js
iruzevic: ok

iruzevic (approved):

volfkarlo:
This should be a minor release. Since it's only a fix and not a feature.
I tagged it with 6.1.0 but I believe it should be 6.0.1.

@iruzevic said we will make a minor release next week so I'll leave it here until then.

dingo-d (approved):

PR #621, opened 2022-05-23
Adds media-breakpoints variable that pulls variables from global manifest.json

moonbyt3:
# Description

Added variable $media-breakpoints to _shared-variables.scss

$media-breakpoints is defined in node_modules>media-blender with !default keyword which means if there is no variable defined, use these values.
I had to reassign that variable in eightshift-frontend-libs and pull data from blocks manifest.json with global-settings function.


mbmjertan (approved):

piqusy:
Would it be over-engineering to create a SCSS function which takes `global-settings(breakpoints)` as a parameter from global manifest and outputs all needed breakpoints?

iruzevic (approved):

PR #625, opened 2022-05-25
WIP: adding new featured content block

iruzevic:
⚠️ DO NOT REVIEW YET

# Description

- new featured content block.
- new layout component with 2-3-4 columns

TODO:

- [ ] featured content block
- [ ] layout component
- [ ] load more component
- [ ] setup featured content preview on storybook


blocks/init/src/Blocks/components/load-more/docs/readme.mdx
```diff
@@ -0,0 +1,139 @@
...
+4. You must provide load more component `loadMoreQuery` attribute with encoded json string with all WP_Query args you want to use. In general this is the same query you use to provide the initial data.
+5. You must provide load more component `loadMoreType` attribute that will determin what map you will use in you dynamic data class to provide view output.
+5. You must provide load more component `loadMoreInitiaItems` attribute that contains the array of IDs for the initial loaded data in the DOM.
+6. Optional load more component `loadMorePerPageOverride` attribute is used to load different number of items than they are initialy set in the DOM.
+7. After all this items are connected and provided, when you click on the load more button it will make a request using JS on the admin ajax hook in your dynamic data class.
+8. Dynamic data will provide the query data that you will map to your component view and provide the data back to the dom via ajax.
+9. In you component you must have partial view component that will map ids from the dynamic data to the actial output.
```
iobrado, line 35:
```suggestion
9. In your component you must have partial view component that will map ids from the dynamic data to the actual output.
```

blocks/init/src/Blocks/components/load-more/load-more.php
```diff
@@ -0,0 +1,64 @@
...
+$blockClass = $attributes['blockClass'] ?? '';
+$selectorClass = $attributes['selectorClass'] ?? $componentClass;
+
+$loadMoreInitiaItems = Components::checkAttr('loadMoreInitiaItems', $attributes, $manifest, $componentName);
+$loadMoreQuery = Components::checkAttr('loadMoreQuery', $attributes, $manifest, $componentName);
+$loadMoreId = Components::checkAttr('loadMoreId', $attributes, $manifest, $componentName);
+$loadMoreType = Components::checkAttr('loadMoreType', $attributes, $manifest, $componentName);
+$loadMorePerPageOverride = Components::checkAttr('loadMorePerPageOverride', $attributes, $manifest, $componentName);
+
+// Bailout if no query is not provided.
```
iobrado, line 31:
```suggestion
// Bailout if no query is provided.
```

blocks/init/src/Blocks/custom/featured-content/components/featured-content-options.js
```diff
@@ -0,0 +1,237 @@
...
+	const getTaxonomyEndpoint = taxonomyOptions.filter((item) => item.value === featuredContentTaxonomy)[0]?.api ?? [];
+
+	const WarningInfo = () => {
+		return (
+			<>
+				<Notice
+					status={"warning"}
+					isDismissible={false}
+				>
+					{__('This options can only be used in post single. Keep in mind by selecting this option the editor preview may not show the correct items!', 'eightshift-boilerplate')}
```
iobrado, line 65:
```suggestion
					{__('These options can only be used in post single. Keep in mind by selecting this option the editor preview may not show the correct items!', 'eightshift-boilerplate')}
```

blocks/init/src/Blocks/custom/featured-content/components/featured-content-options.js
```diff
@@ -0,0 +1,237 @@
...
+							simpleValue={true}
+						/>
+					}
+
+					{featuredContentTaxonomy &&
+						<>
+							<IconToggle
+								icon={icons.hoverBackgroundType}
+								label={__('Select specific items', 'eightshift-boilerplate')}
+								help={__('This options will allow you to manually select items to show.', 'eightshift-boilerplate')}
```
iobrado, line 120:
```suggestion
								help={__('These options will allow you to manually select items to show.', 'eightshift-boilerplate')}
```

iobrado (changes requested):
This is awesome! 👏 Left some typo fixes 🙂

blocks/init/src/Blocks/custom/featured-content/featured-content.php
```diff
@@ -0,0 +1,142 @@
...
+		$args['tax_query'][0]['operator'] = 'NOT IN'; // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_tax_query
+	}
+}
+
+$excludeList = [];
+
+if ($featuredContentExcludeCurrentPost && $post instanceof WP_Post) {
+	$excludeList[] = $post->ID;
+}
+
```
gabriel-glo, line 66: It would be much better if we could have this inside a class method. Ideally this will be a filter that takes multiple arguments (as WP_Query args) and then just returns the query object in `$mainQuery` variable.

blocks/init/src/Blocks/custom/featured-content/featured-content.php
```diff
@@ -0,0 +1,142 @@
...
+if ($excludeList) {
+	$args['post__not_in'] = $excludeList;
```
gabriel-glo, line 68: This should be handled inside the WP_Query loop. We'd check if current post ID is same as the one in the loop and just `continue;` with the loop

gabriel-glo (changes requested):
A couple of things to consider.
This is so awesome.

blocks/init/src/Blocks/components/layout/docs/readme.mdx
```diff
@@ -0,0 +1,11 @@
+# Layout Component
+
+Component used to provide layout grid columns functionality to your project.
```
dingo-d, line 3:
```suggestion
A component that is used to provide a layout (grid or columns) functionality to your project.
```

This seems a bit more descriptive as towhat the component actually provides (a container wrapper basically, no?).

blocks/init/src/Blocks/components/layout/layout.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+/**
+ * Layout component - Grid Columns grid.
```
dingo-d, line 4: Now I'm unsure about my comment in the storybook readme.

What is this layout component actually doing?

Just providing the CSS grid wrapper? Wouldn't then it be better for it to be named `grid` or `layout-grid`, instead of just layout?

I think the core (or just wp.com) have a Layout Grid block:

image

Layout implies a specific structure

blocks/init/src/Blocks/components/layout/components/layout-options.js
```diff
@@ -0,0 +1,79 @@
...
+		showLayoutTotalItems = true,
+	} = attributes;
+
+	if (!layoutShowControls) {
+		return null;
+	}
+
+	const layoutUse = checkAttr('layoutUse', attributes, manifest);
+	const layoutType = checkAttr('layoutType', attributes, manifest);
+	const layoutTotalItems = checkAttr('layoutTotalItems', attributes, manifest);
```
dingo-d, line 38: This naming is a bit odd, the label says `Maximum items to show`, would it be better to name it `layoutMaxNumberOfItems`?

blocks/init/src/Blocks/components/layout/components/layout-options.js
```diff
@@ -0,0 +1,79 @@
...
+								isActive: layoutType === value,
+								icon: icons[iconName],
+							};
+						})}
+					/>
+				}
+
+				{showLayoutTotalItems &&
+					<RangeControl
+						label={<IconLabel icon={icons.totalItems} label={__('Maximum items to show', 'eightshift-boilerplate')} />}
```
dingo-d, line 68:
```suggestion
						label={<IconLabel icon={icons.totalItems} label={__('Maximum number of items to show', 'eightshift-boilerplate')} />}
```

blocks/init/src/Blocks/components/layout/layout.php
```diff
@@ -0,0 +1,50 @@
...
+$unique = Components::getUnique();
+
+?>
+
+<<?php echo esc_attr($layoutTag); ?>
+	class="<?php echo esc_attr($layoutClass); ?>"
+	data-id="<?php echo esc_attr($unique); ?>"
+	data-layout-type="<?php echo esc_attr($layoutType); ?>"
+>
+	<?php echo Components::outputCssVariables($attributes, $manifest, $unique); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>
```
dingo-d, line 43: You shouldn't need to add the `// phpcs:ignore` comment here, this is a safe method

https://github.com/infinum/eightshift-coding-standards/blob/main/Eightshift/Sniffs/Security/ComponentsEscapeSniff.php#L146-L147

blocks/init/src/Blocks/components/load-more/assets/index.js
```diff
@@ -0,0 +1,25 @@
...
+
+	const selectors = `.${componentJsClass}`;
+	const elements = document.querySelectorAll(selectors);
+
+	if (elements.length) {
+		import('./load-more').then(({ LoadMore }) => {
+			const loadMore = new LoadMore({
+				triggerElements: elements,
+				ajaxHandler: 'dynamic_data',
+				ajaxUrl: esBlocksLocalization?.ajaxurl,
```
dingo-d, line 19: Why are we using AJAX for load more functionality instead of using REST API?

REST API is better suited for this (security, performance, etc.)

blocks/init/src/Blocks/components/load-more/assets/load-more.js
```diff
@@ -0,0 +1,322 @@
...
+		const data = this.getData(id);
+
+		// Bailout if there are no more actions.
+		if (data.preventAction) {
+			return;
+		}
+
+		// Prepare body data.
+		const body = {
+			method: 'POST',
```
dingo-d, line 151: _Fetching_ implies `GET`, not `POST` 😬 

We should use appropriate HTTP methods for things we do with the API (or AJAX in this case).

blocks/init/src/Blocks/components/load-more/load-more.php
dingo-d: Or

```suggestion
// Bailout if the query is not provided.
```

?

blocks/init/src/Blocks/custom/featured-content/components/featured-content-options.js
dingo-d, line 65:
```suggestion
					{__('These options can only be used in a single post. Keep in mind that, by selecting this option, the editor preview may not show the correct items!', 'eightshift-boilerplate')}
```

blocks/init/src/Blocks/custom/featured-content/components/featured-content-options.js
```diff
@@ -0,0 +1,237 @@
...
+								/>
+							}
+						</>
+					}
+				</>
+			}
+
+			<IconToggle
+				icon={icons.none}
+				label={__('Exclude current post', 'eightshift-boilerplate')}
```
dingo-d, line 203:
```suggestion
				label={__('Exclude the current post', 'eightshift-boilerplate')}
```

blocks/init/src/Blocks/custom/featured-content/components/featured-content-options.js
```diff
@@ -0,0 +1,237 @@
...
+				checked={featuredContentExcludeCurrentPost}
+				help={__('Exclude current post from the featured content block items.', 'eightshift-boilerplate')}
```
dingo-d, line 205:
```suggestion
				help={__('Exclude the current post from the featured content block items.', 'eightshift-boilerplate')}
```

blocks/init/src/Blocks/custom/featured-content/featured-content.php
dingo-d: 💯

blocks/init/src/Blocks/custom/featured-content/featured-content.php
```diff
@@ -0,0 +1,142 @@
...
+	<?php
+	echo Components::outputCssVariables($attributes, $manifest, $unique);
+
+	$cards = Components::render(
+		'cards.php',
+		[
+			'items' => $mainQuery->posts,
+			'blockSsr' => $featuredContentServerSideRender,
+		],
+		__DIR__ . '/partials',
```
dingo-d, line 112: Will this be an issue on WIN systems (directory separator)?

blocks/init/src/Blocks/custom/featured-content/manifest.json
```diff
@@ -0,0 +1,79 @@
...
+					"category"
+				]
+			},
+			{
+				"label": "Page",
+				"value": "page",
+				"api": "pages"
+			},
+			{
+				"label": "News",
```
dingo-d, line 66: This is not a default post type, that could cause an issue, no?

dingo-d (changes requested):
Left some questions and suggestions. Like the direction where this is going, but the PR is encompassing a bit more than just a featured content block. 

For instance, I'd like for the layout component to be extracted in a separate PR for easier review.

You can always create branches, cherry pick, create a separate PR, then once that is merged to develop branch, just rebase the current PR. Keeps things a bit easier to review 😉

blocks/init/src/Blocks/components/layout/components/layout-options.js
iruzevic: i like the keep the attribute naming as short and simple as possible. The label is here for the users. So i think this one is ok

blocks/init/src/Blocks/components/layout/layout.php
iruzevic: the idea here is to have any kind of layouts here not just grid like. So this is why I just called it layout. I will update the readme

blocks/init/src/Blocks/components/layout/layout.php
iruzevic: tnx. I have fixed it on all components

blocks/init/src/Blocks/components/load-more/assets/load-more.js
iruzevic: The problem here is that I send the data from frontend to backend via json format because we have multiple decoded params that can't be sent via get parameter. So this is why I was using POST

blocks/init/src/Blocks/custom/featured-content/featured-content.php
iruzevic: agreed but this will overcomplicate the implementation so I would like to not do that.

blocks/init/src/Blocks/custom/featured-content/featured-content.php
iruzevic: agreed, but with this approach pagination is a nightmare so In general there max 1-2 items here if any so lets not overcomplicate this for a minor optimisation.

dadadavorin (approved):
Awesome! 🤩

blocks/init/src/Blocks/custom/featured-content/featured-content.php
dingo-d: We'll need to really be careful about this because there are projects that have over 10k posts, and this could be a bottleneck.

blocks/init/src/Blocks/components/load-more/assets/load-more.js
dingo-d: I think we discussed this yesterday, no?

dingo-d:
Once the tests are written just squash merge this to have a cleaner commit history.

blocks/init/src/Blocks/custom/featured-content/featured-content.php
iruzevic: i have opened an issue https://github.com/infinum/eightshift-frontend-libs/issues/631

blocks/init/src/Blocks/custom/featured-content/featured-content.php
iruzevic: agreed, i have created an issue for this https://github.com/infinum/eightshift-frontend-libs/issues/632

PR #642, opened 2022-07-04
Add check for forbidden keywords

JerrySpecter:
# Description

Add check for forbidden keywords. If the `projectName` argument gets a forbidden keyword it trows an error and prompts you to enter another name.

Example with CLI argument:
image

Example without arguments:
image


iruzevic (approved):
looks good

iruzevic (changes requested):
we have tests in JS so please make this test cases

package.json
```diff
@@ -64,6 +64,7 @@
 		"file-loader": "^6.2.0",
 		"husky": "^7.0.4",
 		"import-glob-loader": "^1.1.0",
+		"inquirer": "^9.0.0",
```
iruzevic, line 67: why is this package installed ? what is it used for?

package.json
```diff
@@ -95,6 +96,7 @@
 		}
 	},
 	"devDependencies": {
+		"@babel/plugin-transform-modules-commonjs": "^7.18.6",
```
iruzevic, line 99: commnet from babel

iruzevic (changes requested):

JerrySpecter:
Pushed new changes, added tests, removed unused dependencies. Locally tests pass successfully, is there any way i can recreate Jest errors show here?

PR #647, opened 2022-07-14
Add checks for requirements

JerrySpecter:
# Description
Add checks for requirements before theme setup starts.

System requirements check passed successfully:
image

Check fails if composer is not installed:
image

latest updated:
image

### DIFF
--- a/setup/create-wp-project/src/basics/argument-operations.js
+++ b/setup/create-wp-project/src/basics/argument-operations.js
@@ -14,6 +14,7 @@ const summary = async(answers) => {
     log(`- ${key}: ${variable(answers[key])}`);
   });
   
+  // Write out all the requirements needed to run the script without any errors
   log('');
   log(label('Requirements: '));
   log('Please check if you have installed these packages before you start the setup:');
--- a/setup/create-wp-project/src/basics/argument-operations.js
+++ b/setup/create-wp-project/src/basics/argument-operations.js
@@ -13,7 +13,20 @@ const summary = async(answers) => {
   Object.keys(answers).forEach((key) => {
     log(`- ${key}: ${variable(answers[key])}`);
   });
-
+  
+  log('');
+  log(label('Requirements: '));
+  log('Please check if you have installed these packages before you start the setup:');
+  log(`${variable('1.')} Node.js - LTS`);
+  log(`- check by running: ${variable('node -v')}`);
+  log(`${variable('2.')} Composer - LTS`);
+  log(`- check by running: ${variable('composer -v')}`);
+  log(`${variable('3.')} WP-CLI - LTS`);
+  log(`- check by running: ${variable('wp --info')}`);
+  log(`${variable('4.')} Git - LTS`);
+  log(`- check by running: ${variable('git --version')}`);
+  log('');
+  
   const { confirmSummary } = await inquirer.prompt({
     name: 'confirmSummary',
     type: 'confirm',
--- a/setup/create-wp-project/src/basics/argument-operations.js
+++ b/setup/create-wp-project/src/basics/argument-operations.js
@@ -22,9 +22,9 @@ const summary = async(answers) => {
   log(`- check by running: ${variable('node -v')}`);
   log(`${variable('2.')} Composer - LTS`);
   log(`- check by running: ${variable('composer -v')}`);
-  log(`${variable('3.')} WP-CLI - LTS`);
+  log(`${variable('3.')} WP-CLI`);
   log(`- check by running: ${variable('wp --info')}`);
-  log(`${variable('4.')} Git - LTS`);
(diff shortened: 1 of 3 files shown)


setup/create-wp-project/src/basics/command-line.js
```diff
@@ -5,6 +5,27 @@ const { exec } = require('promisify-child-process');
...
 const cloneTimeout = 45000;
 
+const checkNodeVersion = async () => exec(`node -v`);
+const checkComposerVersion = async () => exec(`composer -v`);
+const checkWPCliVersion = async () => exec(`wp --info`);
+const checkGitVersion = async () => exec(`git --version`);
+
+const checkRequirements = async () => {
+  const nodeVersion = await checkNodeVersion();
+  if (nodeVersion.stderr !== '') process.exit(0);
```
iruzevic, line 15: will the user get the somekind of msg if the node is not instaled?

JerrySpecter: If you don't have node installed, the script can't start. if you trigger `npx create-wp-project` or if you try to manually trigger the script with for example `node  ~/infinum/eightshift-frontend-libs/setup/create-wp-project/create-wp-project.js` without node installed you will get a message saying `command not found: npx` or `command not found: node`

iruzevic: is there a way to add some kind of user friendly msg other than `command not found: node`?

mbmjertan: Just for clarity: if you have `npx` installed, you'll have node installed as well. You also won't be able to run this script if you don't have node installed (not taking into account Node-compatible runtimes, which are rare, but exist). I expect this only to fail in cases where `node` isn't in a user's `$PATH`.

As discussed below:

```suggestion
  if (nodeVersion.stderr !== '') process.exit(1);
```

setup/create-wp-project/src/basics/command-line.js
```diff
@@ -5,6 +5,27 @@ const { exec } = require('promisify-child-process');
...
+  
+  const composerVersion = await checkComposerVersion();
+  if (composerVersion.stderr !== '') process.exit(0);
```
mbmjertan, line 18: None of these should exit with the status `0`, as status code 0 indicates successful execution:

> A successful command returns a 0, while an unsuccessful one returns a non-zero value that usually can be interpreted as an error code. Well-behaved UNIX commands, programs, and utilities return a 0 exit code upon successful completion, though there are some exceptions.

(from https://tldp.org/LDP/abs/html/exit-status.html)

Other scripts, such as deploy agents or provisioning scripts could rely on `npx create-wp-project`, so we should take care that it plays nicely.

```suggestion
  if (composerVersion.stderr !== '') process.exit(1);
```

setup/create-wp-project/src/basics/command-line.js
```diff
@@ -5,6 +5,27 @@ const { exec } = require('promisify-child-process');
...
+  
+  const wpCliVersion = await checkWPCliVersion();
+  if (wpCliVersion.stderr !== '') process.exit(0);
```
mbmjertan, line 21:
```suggestion
  if (wpCliVersion.stderr !== '') process.exit(1);
```

setup/create-wp-project/src/basics/command-line.js
```diff
@@ -5,6 +5,27 @@ const { exec } = require('promisify-child-process');
...
+  
+  const gitVersion = await checkGitVersion();
+  if (gitVersion.stderr !== '') process.exit(0);
```
mbmjertan, line 24:
```suggestion
  if (gitVersion.stderr !== '') process.exit(1);
```

mbmjertan (changes requested):
Some comments regarding status codes, otherwise looks good! 😄

setup/create-wp-project/src/basics/argument-operations.js
```diff
@@ -13,7 +13,20 @@ const summary = async(answers) => {
...
+  log('Please check if you have installed these packages before you start the setup:');
+  log(`${variable('1.')} Node.js - LTS`);
+  log(`- check by running: ${variable('node -v')}`);
+  log(`${variable('2.')} Composer - LTS`);
+  log(`- check by running: ${variable('composer -v')}`);
+  log(`${variable('3.')} WP-CLI - LTS`);
+  log(`- check by running: ${variable('wp --info')}`);
+  log(`${variable('4.')} Git - LTS`);
+  log(`- check by running: ${variable('git --version')}`);
+  log('');
```
mbmjertan, line 29: There are no LTS versions of WP CLI and Git 😄

JerrySpecter: Removed the version flags

PR #727, opened 2023-05-22
Feature/wp cli rename dummy block

moonbyt3:
# Description

Created two new folders, one in blocks and one in components, called dummy. Files in these folders have incorrect syntax due to placeholders that we use to rename stuff. I've also added phpcs and phpstan ignores for these folders so linters don't whine.


goranalkovic-infinum:
Could you maybe prefix the dummy block folders with `_`? It's highly unlikely that someone will name their block _dummy_, but you never know 😄

iruzevic:
It will be by the end of the year

PR #820, opened 2024-04-25
Install Theme no matter if the user run the npx create-wp-project command in the wrong folder

DjidjaleskaSandra:
# Description
The task aims to create a Node.js function that identifies the root directory of a WordPress project by searching for the presence of the wp-config.php file. This functionality will facilitate automated processes within the project setup or maintenance scripts, ensuring they execute from the correct directory.

# Screenshots / Videos
https://github.com/infinum/eightshift-frontend-libs/assets/97440314/7c1c7af7-cf2e-4f3e-82af-d0ec79a4c18b

# Productive
https://app.productive.io/1-infinum/tasks/7791999


setup/create-wp-project/src/commands/theme.js
```diff
@@ -36,6 +36,15 @@ exports.handler = async (argv) => {
 	await clearConsole();
 	await writeIntro();
 
+	// Check if you are in the required folder.
+	const { path: requiredPath, isCorrectFolder } = await installPath('themes');
+
+	if (!isCorrectFolder) {
```
iruzevic, line 42: The goal is to allow users to set up the project from the root directory, without being required to be inside the theme folder. The installation process will ensure that the theme is placed in the correct location.

So if they setup the project from the root, wp-content, wp-content/plugins no mater the setup should go to the root of the project where the WP-config is set add add the theme to the correct place.

The same thing should work with the plugin flag used for the plugin setup

iruzevic (changes requested):

setup/create-wp-project/src/commands/theme.js
```diff
@@ -36,6 +36,39 @@ exports.handler = async (argv) => {
 	await clearConsole();
 	await writeIntro();
 
+	const fs = require('fs');
+
+	const findProjectRoot = (currentDir) => {
```
iruzevic, line 41: how will this work with the plugin flag?

iruzevic (changes requested):

setup/create-wp-project/src/commands/theme.js
DjidjaleskaSandra: I think Plugins to be in a separate task, but putting everything together into one PR can also work, especially when the changes are closely connected and won't get confusing. So, we can go ahead with that approach :).

iruzevic (approved):
looking good, please wait for the review from @goranalkovic-infinum

iruzevic:
I have used parts of your PR for the new release so I will not merge this but parts of this is in the new release. tnx

PR #823, opened 2024-04-30
release 11.0.0

iruzevic:
Dependant to [Eightshift-libs@8.0.0](https://github.com/infinum/eightshift-libs/releases/tag/8.0.0)

### Changed
- All `Components::getManifest` to `Components::getManifestByDir` method.
- All `Components::outputCssVariables($attributes, $manifest, $unique, $globalManifest)` to `Components::outputCssVariables($attributes, $manifest, $unique)`.
- All `renderPartial` to `render` method.
- Wrapper changes `$this->renderWrapperView` to `echo $renderContent;`.

## Todo:

- [ ] make migration guide
- [ ] test on a bigger project
- [x] write changelog
- [ ] make a release


blocks/init/src/Blocks/components/button/button.php
```diff
@@ -62,7 +61,7 @@
 ?>
 
 <?php
-	echo Components::outputCssVariables($attributes, $manifest, $unique, $globalManifest);
+	echo Components::outputCssVariables($attributes, $manifest, $unique);
```
goranalkovic-infinum, line 64: Shouldn't these be `Helpers::` ?

goranalkovic-infinum (changes requested):

blocks/init/src/Blocks/components/button/button.php
iruzevic: yes, it should, but I will make this a separate PR because there are a lot of changes.

PR #824, opened 2024-04-30
Make installation available from other branches than the Infinum repo

### LINKED ISSUE #749: [FEATURE] - Make installation available from other branches than the infinum repo
Currently, the `eightshiftLibsRepo` is a hardcoded value, which makes installation of different forks for testing purposes, a bit harder (you need to manually change repos). Would it be possible to make this configurable as well? By default, the branch will be the infinum's repo, but by passing the correct repo info, it would be pulled from that repo instead.

https://github.com/infinum/eightshift-frontend-libs/blob/25e68168ed5c8247fd134ff0df762c757cd9e2d5/setup/create-wp-project/src/dependencies.js#L34-L36

DjidjaleskaSandra:
# Description

This PR introduces functionality that enables users to specify custom repository URLs for Composer dependencies directly via command line arguments. This enhancement provides greater flexibility and control over dependency management, particularly useful for testing different branches or forks of libraries during development.

# Productive ticket

https://app.productive.io/1-infinum/tasks/7821114


iruzevic:
Looks ok, but why not add frontend libs and boilerplate repos flags also?

DjidjaleskaSandra:
> Looks ok, but why not add frontend libs and boilerplate repos flags also?

Bolilerplate repo -flag exist, I added for frontend libs only

iruzevic (approved):

iruzevic:
I have changed the new setup full so the part of these changes are there I will not be merging the PR but tnx

PR #847, opened 2025-01-28
Add Secure and SameSite parameters as options to cookie method

dadadavorin:
# Description

We started getting errors from BugSnag for unsecured cookies, so I solved this on a project, but consider we should also have this as an option in FE Libs cookie helper.

This will set `sameSite` default to `Lax` (which is already default, but not in all browsers, like Safari or Firefox). Means that cookie is not sent on cross-site requests by default. More [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#lax).

Also, set `Secure` attribute by default, which indicates that the cookie is sent to the server only when a request is made with the https: scheme (except on localhost), and therefore, is more resistant to [man-in-the-middle](https://developer.mozilla.org/en-US/docs/Glossary/MitM) attacks.

### DIFF
--- a/scripts/helpers/cookies.js
+++ b/scripts/helpers/cookies.js
@@ -23,7 +23,7 @@ export const cookies = {
 	 * cookies.setCookie('gdpr', '2', cookies.setOneDay(), '/', '.example.com', true, 'Strict');
 	 * ```
 	 */
-	setCookie(key, value, time, path, domain, secure = false, sameSite = 'Lax') {
+	setCookie(key, value, time, path, domain, secure = true, sameSite = 'Lax') {
 		const expires = new Date();
 		expires.setTime(expires.getTime() + (time));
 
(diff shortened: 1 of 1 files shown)


scripts/helpers/cookies.js
```diff
@@ -35,7 +38,27 @@ export const cookies = {
...
 		}
 
-		document.cookie = `${key}=${value}${pathValue}${domainValue};expires=${expires.toUTCString()}`;
+		const cookieParts = [
+			`${key}=${value}`,
+			`expires=${expires.toUTCString()}`,
+			`SameSite=${sameSite}`,
+			pathValue,
+			domainValue,
+		];
```
goranalkovic-infinum, line 47: Please rewrite this so it's an object instead of array of strings

goranalkovic-infinum (changes requested):

iruzevic (approved):

PR #851, opened 2025-04-24
Release/14 update

iruzevic:
### Changed

- Webpack config now user mjs files and no longer has assets config key.
- Es and style linters

### Removed

- Most of the SCSS mixing and files
- All initial Blocks


package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@eightshift/frontend-libs",
-	"version": "14.0.1",
+	"version": "14.0.2",
```
goranalkovic-infinum, line 3:
```suggestion
	"version": "15.0.0",
```

goranalkovic-infinum (changes requested):

package.json
iruzevic: you are not giving this up :P

package.json
```diff
@@ -25,7 +25,7 @@
 	},
 	"scripts": {
 		"lintStyle": "stylelint **/*.scss",
-		"lintJs": "eslint blocks/ webpack/ scripts/ linters/",
+		"lintJs": "npx eslint",
```
goranalkovic-infinum, line 28:
```suggestion
		"lintJs": "bunx eslint",
```

PR #861, opened 2026-05-14
feature/improvements

iruzevic:
# Description

Release 17.0.0 — dependency refresh, performance refinements, and a couple of breaking changes around Webpack defaults.

### Added

- Added `eslint-plugin-unused-imports` to the base ESLint config to flag unused imports and variables.
- Added `extraEntries` and `extraPlugins` options to the Webpack project config so projects can extend entry points and plugins without overriding the full config.
- Added `at-rule-no-unknown` to the Stylelint config with the SCSS at-rule allowlist to surface unknown at-rules without breaking SCSS files.

### Changed

- Updated all dependencies to their latest versions.
- Updated React ESLint setting from a pinned version `18` to `detect`.
- Updated `store.js` reducer cases (`SET_STYLE`, `SET_STYLE_BY_INDEX`, `UNSET_STYLE_BY_INDEX`) to return new state objects instead of mutating existing state.
- Updated `getCookie` helper to parse cookies via `document.cookie.split` and `decodeURIComponent` for safer value handling.
- Updated `attributes.js`, `options.js`, and related helpers with `WeakMap`-based caching for `componentName` lookups and other performance refinements.
- Updated `webpack/helpers.mjs` to expose `readManifestJson` and `convertDataToSass` helpers, reducing duplicate manifest reads.

### Removed

- Removed the global `jQuery` and `$` Webpack `ProvidePlugin` injection. Projects that relied on globally injected jQuery must import it explicitly.
- Removed the `process.browser` Webpack `DefinePlugin` value in favor of `typeof window: '"object"'`. Code that referenced `process.browser` must be updated.
- Removed unused `React` import from `css-variables.js` and cleaned up other unused imports across components.

### Fixed

- Fixed the MJS Webpack rule regex (`/\.m?js/` → `/\.m?js\$/`) so non-JS paths containing `js` are no longer matched.
- Downgraded ESLint to `^9.39.4` to restore compatibility with the current plugin ecosystem.

## QA Guide

Steps for QA to test this feature:
1. Install \`@eightshift/frontend-libs@17.0.0\` in a test project and run a full Webpack build.
2. Verify SCSS files compile and \`@use\`/\`@forward\`/\`@mixin\` etc. don't trigger Stylelint errors.
3. Run \`eslint\` against the project — unused imports should be flagged as errors.
4. Confirm blocks register and the editor store updates styles without console warnings.
5. Verify cookies set with special characters (URI-encoded) read back correctly via \`getCookie\`.
6. Confirm any code that previously relied on global \`jQuery\` / \`$\` is updated to import jQuery explicitly.

**Expected result:** Build succeeds, linters run, blocks render in the editor, styles update reactively, and there are no console errors related to undefined \`jQuery\` or \`process.browser\`.

# Screenshots / Videos

# Linked documentation PR


package.json
```diff
@@ -37,41 +37,42 @@
...
-		"browserslist": "^4.28.1",
+		"@stylistic/stylelint-plugin": "^5.1.0",
+		"@swc/core": "^1.15.33",
+		"@wordpress/dependency-extraction-webpack-plugin": "^6.45.0",
+		"autoprefixer": "^10.5.0",
+		"browserslist": "^4.28.2",
 		"css-loader": "^7.1.4",
 		"css-minimizer-webpack-plugin": "^8.0.0",
-		"eslint": "^10.0.3",
+		"eslint": "^9.39.4",
```
goranalkovic-infinum, line 47: Which plugin isn't compatible with v10?

.github/workflows/ci.yml
```diff
@@ -22,11 +22,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node: [ 'latest' ]
+        node: ['latest']
```
goranalkovic-infinum, line 25:
```suggestion
        node: ['24']
```
We should probably keep it at LTS

scripts/components/toggle/toggle.js
```diff
@@ -22,7 +22,7 @@ export const Toggle = ({
 	type = 'toggle',
 
 	help,
-	inlineHelp = false,
+	inlineHelp: _inlineHelp = false,
```
goranalkovic-infinum, line 25: If this is not used anywhere, you can remove it

goranalkovic-infinum (changes requested):

.github/workflows/ci.yml
iruzevic: stupid AI i will revert

